PR #2, opened 2020-08-03
Add color values and ability to setup with highlight groups

Akin909:
This PR adds colors for the devicons as well as creating highlight groups if the `setup` function is called.

A couple of things to note:
1. The colours are largely from `github` (github/linguist) ultimately but via stealing the json from `chadtree.vim` where these colors are stored in json.

2. I made the logic dumber but simpler and I hope less potentially error prone i.e. rather than using a regex to derive the names for a highlight group, I just manually (vim macro) gave each group a name. This means we just check the name and concatenate `DevIcon` and use that as the name for the highlight group.

3. I've also added a few more icons.

Example result using api in `nvim-bufferline.lua`

colors-nvim-devicons

NOTE: Tried this with `nvim-tree.lua` and nothing seems to be broken.


Akin909:
### Open question
`vim-devicons` returns a default icon `` if there is none for a filetype. I personally think this is more consistent behaviour than returning nil. I can add this here if you agree

kyazdani42:
thanks, we'll see if it breaks something but i think it should be allright :)

PR #4, opened 2020-08-06
Add an options table so we can choose to return a default

Akin909:
@kyazdani42 this PR changes the way we handle the default icon, it adds in a 3rd optional argument to `get_icon` which are options. For now the only option is `default = boolean` if true then the get icon will return the default icon if none if found. This way a user can still opt not to return a default.

so usage will be
```lua
-- No default
require'nvim-web-devicons'.get_icon("filename.js", "js")

-- Or with default
require'nvim-web-devicons'.get_icon("filename.js", "js", { default = true })
```

I think this is preferable to just removing it and adding the default to exports because by removing it it means you won't be able to get the `highlight name` for the default icon. The default icon's name is `Default` but the highlight group is added when `get_highlight_name` is called so it's actually `DefaultDevIcon` which someone won't know unless they check the code. If they do hard code this though then if we change the way names are created people using this will have to change their plugins to match the new name.

Also in `bufferline.lua` it means I have to do `nil` checks manually then add and highlight the default icon and group as well

### DIFF
--- a/lua/nvim-web-devicons.lua
+++ b/lua/nvim-web-devicons.lua
@@ -820,8 +820,7 @@ local default_icon = {
 }
 
 return {
-  default_icon = default_icon,
-  get_icon = function(name, ext)
+  get_icon = function(name, ext, opts)
     local icon_data = icons[name]
     local by_name = icon_data and icon_data.icon or nil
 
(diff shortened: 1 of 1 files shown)


kyazdani42:
Right thanks :)
we need to write some docs for this plugin ^^

PR #16, opened 2020-10-25
feat: option to globally enable default icon

elianiva:
This will allow user to enable default icon globally. It will benefit plugins like [telescope](https://github.com/nvim-lua/telescope.nvim) (and some other plugins who uses devicons) so you don't have to configure it for each plugin.

### DIFF
--- a/lua/nvim-web-devicons.lua
+++ b/lua/nvim-web-devicons.lua
@@ -824,6 +824,11 @@ local default_icon = {
   name = "Default",
 }
 
+local global_opts = {
+  override = {},
+  default = false
+}
+
 return {
   get_icon = function(name, ext, opts)
     local icon_data = icons[name]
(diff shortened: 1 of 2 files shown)


lua/nvim-web-devicons.lua
```diff
@@ -833,17 +838,27 @@ return {
       return by_name, get_highlight_name(icon_data)
     else
       icon_data = icons[ext]
-      if opts and opts.default and not icon_data then
+
+      if global_opts and global_opts.default and not icon_data then
+        icon_data = default_icon
+      elseif opts and opts.default and not icon_data then
```
kyazdani42, line 844: you don't need to branch here
```lua
if (global_opts.default or (opts and opts.default)) and not icon_data then
  icon_data = default_icon
end
```

lua/nvim-web-devicons.lua
```diff
@@ -833,17 +838,27 @@ return {
...
     end
   end,
-  setup = function(user_icons)
-    icons = vim.tbl_extend("force", icons, user_icons or {});
+  setup = function(opts)
+    local user_icons = opts and opts.override or {}
+
+    if global_opts and not global_opts.default then
+      global_opts.default = opts and opts.default or false
+    end
```
kyazdani42, line 859: you don't need to check for global_opts existence as it's declared locally already. same for opts.
```lua
local user_icons = opts or {}
if user_icons.default then
  global_opts.default = true
end

icons = vim.tbl_extend("force", icons, user_icons.override or {})

**Review by kyazdani42 — CHANGES_REQUESTED** [review]

**Comment by elianiva** [comment]
ok, I've changed it.

File: lua/nvim-web-devicons.lua [hunk 3]
```diff
@@ -833,17 +838,25 @@ return {
...
         return by_ext, get_highlight_name(icon_data)
       end
     end
   end,
-  setup = function(user_icons)
-    icons = vim.tbl_extend("force", icons, user_icons or {});
+  setup = function(opts)
+    local user_icons = opts or {}
+
+    if global_opts.default then
```
kyazdani42, line 855: i believe this should be user_icons.default to override global_opts right ?

kyazdani42 (changes requested):

lua/nvim-web-devicons.lua
elianiva: oh yeah, my bad. Will fix it in a sec

elianiva: ok done.

kyazdani42 (approved):

kyazdani42:
thanks :)

PR #39, opened 2021-02-10
perf: defer highlighting to reduce first load time

kevinhwang91:
I don’t know what other people think, but I like to start my nvim quickly

### DIFF
--- a/lua/nvim-web-devicons.lua
+++ b/lua/nvim-web-devicons.lua
@@ -833,12 +833,14 @@ local global_opts = {
 local hl_groups = {}
 
 local function get_highlight_name(data)
-  local hl_group = data.name and "DevIcon" .. data.name
-  if hl_group and not hl_groups[hl_group] and data.color then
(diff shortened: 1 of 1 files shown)


kevinhwang91:
By the way, change the color theme will hi clear, we should add `ColorScheme` to fire highlight again :)

kyazdani42 (approved):
this is a good idea thanks for the PR

akinsho:
@kyazdani42 this PR breaks the usage for `telescope` think a lot of users are probably being affected by this and it would be safer to revert for now, till maybe they can update their code

akinsho:
image

akinsho:
Although tbh I don't think this change is the best way to optimise the startup time. This plugin doesn't automatically call `setup` which is what creates all the highlights it's instead up to the consumer to call setup or more likely call `get_icon` which implicitly calls `setup`. Perhaps the better way to do this would be for consuming plugins to defer calling setup if possible

Issue #44, opened 2021-02-26
Dealing with upper and lower cases in extensions


echasnovski:
Some file extensions allow several forms which differ only in letter cases: '.r' and '.R', '.Rmd' and '.rmd', etc. Currently to account for this, manual extra search should be done (like `devicons.get_icon(file_name, file_ext) or devicons.get_icon(string.lower(file_name), string.lower(file_ext), { default = true })`).

Maybe it is a good idea to incorporate some form of case insensitiveness into plugin directly?

kyazdani42:
Maybe an option to specify matching with or without case sensitiveness, the choice should be left to the user.

echasnovski:
Also noticed that this affects your other plugin, 'nvim-tree.lua':
![nvim-tree_icons](https://user-images.githubusercontent.com/24854248/112178531-7d016a00-8c02-11eb-8ac5-ea9e06ebd644.png)

I don't really know much about case-sensitivity in file extensions, but is it a good idea to add 'R' and 'Rmd' extensions (which are "more default" for R-related files) to [icons](https://github.com/kyazdani42/nvim-web-devicons/blob/1fb0962b8c4a217eec8166b03d683aa070115ed7/lua/nvim-web-devicons.lua#L34) object in this plugin?

kyazdani42:
you could add that indeed, make a PR ! Not sure if other filetypes usually allow uppercase letters

msva:
By the way, this issue is also happens when you dealing with files made by people with case insensitive flesystems.
So, I saw same thing with `*.PDF`, `*.ZIP` files and so on.

gegoune:
@alex-courtis Plugin could sanitise input to all lowercase characters. Can you see any downsides? This might be the easiest solution especially for cases mentioned [above](https://github.com/nvim-tree/nvim-web-devicons/issues/44#issuecomment-1275343459).

PR #48, opened 2021-03-23
Add icons for "R" and "Rmd" extensions.

echasnovski:
This is related to #44. Not sure if it closes it.


kyazdani42:
thanks, partly close it but we should keep it if we find more of these

PR #53, opened 2021-04-01
feat: allow importing all icons

JoosepAlviste:
Hey!

I would like to import all of the icons in my configuration. The use case is to create my own highlight groups for the statusline with a background color:

```lua
local icons = require('nvim-web-devicons').get_icons()
for _, icon_data in pairs(icons) do
  if icon_data.color and icon_data.name then
    local hl_group = icon_data.name and 'StatuslineDevIcon' .. icon_data.name
    if hl_group then
      vim.api.nvim_command("highlight! "..hl_group.. " guifg="..icon_data.color.." guibg="..statusline_background)
    end
  end
end
```

What do you think of this? Should I add the `get_icons` function to the docs, or is it okay to keep it a "hidden" feature?

### DIFF
--- a/README.md
+++ b/README.md
@@ -62,3 +62,13 @@ You can check if the setup function was already called with:
 ```lua
 require'nvim-web-devicons'.has_loaded()
 ```
+
+### Get all icons
+
+It is possible to get all of the registered icons with the `get_icons()` function:
+
+```lua
+require'nvim-web-devicons'.get_icons()
+```
+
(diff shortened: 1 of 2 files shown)


kyazdani42:
Hi ! You could add it to the docs if other people want it, also it might help for debugging purposes :)

JoosepAlviste:
Thanks for taking a look! I added a short description for this to the docs.

kyazdani42 (approved):

kyazdani42:
thanks ;)

PR #54, opened 2021-04-04
Avoid highlights being overridden when setting a colorscheme after setup()

JoosepAlviste:
Hey!

Colorschemes usually run `highlight clear` to clear all highlights before applying their own. This means that if the colorscheme is changed *after* this plugin is initialized, all highlight groups would be lost.

The solution would be to also create the highlights as a response to the `ColorScheme` `autocmd`.

Example of the problem:

https://user-images.githubusercontent.com/9450943/113504038-413d9d00-953e-11eb-8f7d-b68c2471dfe1.mov

Does this solution make sense for this plugin? Also, I wasn't sure if I should wrap the `autocmd` in an `augroup` or not in this case. What do you think?

### DIFF
--- a/lua/nvim-web-devicons.lua
+++ b/lua/nvim-web-devicons.lua
@@ -864,6 +864,17 @@ local function get_highlight_name(data)
   return data.name and "DevIcon" .. data.name
 end
 
+local function set_up_highlights()
+  for _, icon_data in pairs(icons) do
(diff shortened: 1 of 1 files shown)


kyazdani42 (approved):
no strong opinion on augroup, merging this as is, thanks :)

PR #77, opened 2021-09-08
Fix bug caused by multiple setup calls clobbering icon map names

njhoffman:
Simple fix that was causing a lot of problems with another plugin, fzf-lua, and probably others.  If setup gets called multiple times, the icon map names can get clobbered and quit matching extensions properly.


kyazdani42 (approved):
good catch

PR #79, opened 2021-09-09
Add methods to enable icon overrides after setup has been called

### LINKED ISSUE #78: bug: Unable to override/add icons after the latest commit
The solution in the latest commit (1a94ff06dfb0a2240f447dd2e351010725b6fd64) does not seem to be good as that will not allow anyone to add/override an icon.

I have [`yamatsum/nvim-nonicons`](https://github.com/yamatsum/nvim-nonicons) installed which overrides some of the icons from `nvim-web-devicons`, which is successful because it calls the `setup` function first. Then, I also have my own setup which adds custom icons (which now fails) and then there's [`tamago324/lir.nvim`](https://github.com/tamago324/lir.nvim) which adds its own folder icon (which also fails).

A solution would be to expose an `override` function which would allow anyone to add/override an icon without doing the `setup` call.

willruggiano:
Fixes #78

### DIFF
--- a/README.md
+++ b/README.md
@@ -72,3 +72,18 @@ require'nvim-web-devicons'.get_icons()
 ```
 
 This can be useful for debugging purposes or for creating custom highlights for each icon.
+
+
+### Set an icon
+
+You can override individual icons with the `set_icon({...})` function:
+
+```lua
+require("nvim-web-devicons").set_icon {
+  zsh = {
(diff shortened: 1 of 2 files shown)


kyazdani42 (approved):
:+1: thanks

kyazdani42:
hum maybe adding an entry info to the readme ?

PR #85, opened 2021-09-28
Base-16 is now avaiable.

FrenzyExists:
Tweaking each individual icon is a pain, so instead, i decided to apply base-16 to color all icons, and allowed users to simply override a global map called `Colors`. Hope this saves some hastle!

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,11 +9,54 @@ colors for each icon.
 
 ## Installation
 
+Using Plug
 ```vim
 Plug 'kyazdani42/nvim-web-devicons'
 ```
 
+Using Packer
+```lua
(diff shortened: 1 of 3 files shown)


kyazdani42 (commented):
I'm unsure about this. Many icons already have specific colors that don't match the base16 here.
I'd like to get more user feedback before any decision is made towards this.

FrenzyExists:
No prob :+1:

alex-courtis:
As a base16 user I like the idea, however agree with @kyazdani42

kyazdani42:
Let's close this, it's a bit old and it would change the display of most user configurations too drastically.

PR #87, opened 2021-09-30
feat: use only the file name tail in function 'get_icon'

### LINKED ISSUE #84: Question: is intended behavior to return custom icon only on exact match?
There are some file names that have special icons, different for the icon related to the file extension. 
Example: the [```.gitlab-ci.yml```](https://github.com/kyazdani42/nvim-web-devicons/blob/ba8a9362c96bc1fe62038640aa8ce902516707e4/lua/nvim-web-devicons.lua#L155) has its own icon, which is different from the icon for [other ```.yml``` files](https://github.com/kyazdani42/nvim-web-devicons/blob/ba8a9362c96bc1fe62038640aa8ce902516707e4/lua/nvim-web-devicons.lua#L895).

That's great! But the custom icon is only returned if I call the ```get_icon``` function with the exact filename. If the special file is under another path, the default icon is returned:
If call ```:lua print(require('nvim-web-devicons').get_icon('.gitlab-ci.yml'))``` I get the special gitlab icon. ***(Great!)***
But if I call ```:lua print(require('nvim-web-devicons').get_icon('otherdir/.gitlab-ci.yml'))``` I get default ```.yml``` icon.

The same happens if I define some special icon by myself, like ```package.json``` or ```.eslintrc.js```.

This can be resolved by using the function ```vim.fn.fnamemodify``` with the flag ':t' to provide ```get_icon``` only the tail of the file path.

I would like to know if this is the **intended behavior**. If it is, I think a tip could be added to the README instructing the user about using ```vim.fn.fnamemodify``` to not miss the special icons.

I notice that using [telescope.nvim](https://github.com/nvim-telescope/telescope.nvim). With the fuzzy finder, the special icons are rendered only if the file is in the vim current directory, otherwise the 'defaults' icons are displayed. I just want to check with you before submitting a PR to telescope about using ```vim.fn.fnamemodify``` to avoid this inconsistency.

saviocmc:
This makes the function ```get_icon``` return icons consistently when the 'name' parameter passed is a path with other directories, not only the file name. the build in function [fnamemodify](https://neovim.io/doc/user/eval.html#fnamemodify()) removes other parts of the path, leaving only the file name before searching for the icon in the icons table.

@kyazdani42 do you have a way of measure the performance hit of calling ```vim.fn.fnamemodify```?

fix #84

### DIFF
--- a/lua/nvim-web-devicons.lua
+++ b/lua/nvim-web-devicons.lua
@@ -1053,6 +1053,7 @@ local function get_icon(name, ext, opts)
     setup()
   end
 
+  name = vim.fn.fnamemodify(name, ':t')
   local has_default = (opts and opts.default) or global_opts.default
   local icon_data = icons[name] or icons[ext] or (has_default and default_icon)
 


kyazdani42:
i guess you could write a little program to test this on a lot of element, measuring time with `vim.loop.hrtime()` or something like that.

alex-courtis:
nvim-tree has profiling: [log.lua](https://github.com/nvim-tree/nvim-tree.lua/blob/56b626266d634fb667eaf93f179b49bc926ffef2/lua/nvim-tree/log.lua)

You could import that file and invoke `setup` passing
```lua
    log = {
      enable = true,
      truncate = true,
      types = {
        profile = false,
      },
    },
```

Profile around as per [runner.lua](https://github.com/nvim-tree/nvim-tree.lua/blob/56b626266d634fb667eaf93f179b49bc926ffef2/lua/nvim-tree/git/runner.lua#L130)

alex-courtis:
Closing due to inactivity. Please reopen / create a new PR if needed.

PR #88, opened 2021-10-03
Make ext not a required arg

uga-rosa:
If there is no ext, it will automatically read from name.

### DIFF
--- a/lua/nvim-web-devicons.lua
+++ b/lua/nvim-web-devicons.lua
@@ -1049,6 +1049,7 @@ local function setup(opts)
 end
 
 local function get_icon(name, ext, opts)
+  ext = ext or name:match("^.*%.(.*)$") or ""
   if not loaded then
     setup()
   end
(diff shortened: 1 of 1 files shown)


kyazdani42:
could you share an example for why this is needed ? I wouldn't like to break existing functionnality

uga-rosa:
This won't break it. It just looks up the extension from the name of the first argument when the second argument is not given. The existing usage will work fine, and it will also work with only one argument, the file name.
I thought it was useless for the user to bother specifying the extension which can be easily obtained from name.

kyazdani42:
does this still manages to get files with multiple dots ? like what happens for .gitlab-ci-.yml or mix.lock ?

uga-rosa:
The characters after the last dot is used as extension. In this example, it is yml. Well, .gitlab-ci.yml returns a different icon than a normal yml file, so it's not a good example... (I got a little confused during testing).
I appear mix.lock has single dot. These special files do not use extension, so they are not affected by this PR.

uga-rosa:
You can see that by running this snippet.
```lua
print((".hoge.huga.foo"):match("^.*%.(.*)$"))
```
foo will be displayed.

kyazdani42:
alright then, if it causes breakage we'll revert anyway but it should be allright :)

PR #124, opened 2022-04-21
feat: add fennel icon

### LINKED ISSUE #123: Add icon for fennel
https://fennel-lang.org/

vinibispo:
Fix #123

### DIFF
--- a/lua/nvim-web-devicons.lua
+++ b/lua/nvim-web-devicons.lua
@@ -505,6 +505,12 @@ local icons = {
     cterm_color = "185",
     name = "Favicon",
   },
+  ["fnl"] = {
+    color = "#fff3d7",
+    icon = "🌜",
+    cterm_color = "230",
+    name = "Fennel"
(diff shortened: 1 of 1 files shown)


kyazdani42 (approved):

randomizedthinking:
One question. Which font set has the defined icon for fennel?

I tried both DejaVuSansMono and Inconsolata Nerd Font, but none has it. Also, the devicons package does not have an icon for fennel yet.

alex-courtis:
> One question. Which font set has the defined icon for fennel?
> 
> I tried both DejaVuSansMono and Inconsolata Nerd Font, but none has it. Also, the devicons package does not have an icon for fennel yet.

Interesting. It doesn't look like nerd font has it.

Are there any other icons you can't see?
```sh
cd /path/to/nvim-web-devicons
cat lua/nvim-web-devicons.lua | grep -B 1 "^    icon ="
```

PR #132, opened 2022-05-20
Add icons for HDL (hardware description language)

delafthi:
Added icons for:
- vhdl (.vhd, .vhdl)
- verilog (.v, .vh)
- systemverilog (.sv, .svh)

The icon color is debatable. I just took the same green color which is used for the vim icon.


kyazdani42 (approved):
thanks :)

PR #144, opened 2022-07-12
feat(lua): Change deault icon & its color

NikolaM-Dev:


kyazdani42 (commented):
why this change ? you should override this in your configuration i believe.

PR #159, opened 2022-09-23
Icons: Add GraphQL support

luhagel:
So, this isn't quite ideal since nerd fonts don't include the GraphQL logo, but i was tired of seeing the default icon for my gql file and decided to do something about it.

This solution is basically just a recoloured react icon, but given that they are reasonably close you actively have to look to tell, which shouldn't much of an issue on reasonable font sizes.

If you don't think this should be core, I'll gladly leave it as a local tweak in my config :)

### DIFF
--- a/lua/nvim-web-devicons.lua
+++ b/lua/nvim-web-devicons.lua
@@ -596,6 +596,12 @@ local icons = {
     cterm_color = "66",
     name = "GodotProject",
   },
+  ["graphql"] = {
+    icon = "",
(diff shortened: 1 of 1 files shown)


T-727:
I think `nf-fa-connectdevelop` could be a good substitute for GraphQL

luhagel:
@T-727 just updated the pr 👍

kyazdani42 (approved):

PR #170, opened 2022-10-23
Utilize nvim api more

### LINKED ISSUE #169: checking cterm color pollutes `redir` command
Hello, I've noticed that `nvim-web-devicons` `get_icon_*` pollutes vim command output with what seems to be a spurious execution of `:hi <hl_group>` with no arguments. This causes problems when using the `:redir` command because `get_icon_*` functions are usually called by statusline plugins, which fire while `:redir` is registering the output of a command.

## steps to reproduce:

```vim
:redir @a
:lua =require("nvim-web-devicons").get_icon_color('test.py', 'py') 
:redir END
:echo @a

```
produces
```
DevIconPy      xxx ctermfg=61 guifg=#ffbc03
""    "#ffbc03"
```

## expected behaviour

the snippet above should only produce

```
""    "#ffbc03"
```

## solution

I think this is the offending line

https://github.com/nvim-tree/nvim-web-devicons/blob/9061e2d355ecaa2b588b71a35e7a11358a7e51e1/lua/nvim-web-devicons.lua#L1656

I think cterm color could be retrieved using, for example, `nvim_get_hl_by_name('DevIconPy', false).foreground`

T-727:
- Utilizes nvim api more as it is cleaner and more performant
- also localizes some frequently api functions

closes #169

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,7 +24,7 @@ use 'nvim-tree/nvim-web-devicons'
 ### Setup
 
 This adds all the highlight groups for the devicons
-i.e. it calls `highlight IconNameDevicon guifg='<color>' ctermfg='<cterm_color>'` for all icons
+i.e. it calls `vim.api.nvim_set_hl` for all icons
 this might need to be re-called in a `Colorscheme` to re-apply cleared highlights
 if the color scheme changes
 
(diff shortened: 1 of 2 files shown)


alex-courtis:
Thanks for the contribution. I'll soak this for ~week of regular usage.

PR #174, opened 2022-11-14
Add terraform icons

and3rson:
This PR adds icons for `.tf` and `.tfvars`.

I considered `nf-fae-mountains ()`  vs `nf-mdi-terrain (行)` but decided to use the former since it's only 1 character wide.

image

### DIFF
--- a/lua/nvim-web-devicons.lua
+++ b/lua/nvim-web-devicons.lua
@@ -1229,6 +1229,18 @@ local icons = {
     cterm_color = "58",
     name = "Tex",
   },
+  ["tf"] = {
+    icon = "",
+    color = "#5F43E9",
+    cterm_color = "57",
+    name = "Terraform",
+  },
+  ["tfvars"] = {
+    icon = "",
+    color = "#5F43E9",
+    cterm_color = "57",
+    name = "TFVars",
+  },
   ["toml"] = {
     icon = "",
     color = "#6d8086",


alex-courtis (approved):
Thank you for your contribution. Your timing is excellent as I am doing some work with terraform right now...

lanox:
Thanks for this, I was wondering is there reason why we didn't go with normal terraform icons, like the ones that vscode uses?

alex-courtis:
> Thanks for this, I was wondering is there reason why we didn't go with normal terraform icons, like the ones that vscode uses?

Is there a registry of "normal icons"? It might be nice to sync with standards...

T-727:
Assuming I understood correctly, @lanox
This plugin relies on [nerd fonts](https://www.nerdfonts.com/) for it's icons, if they don't provide an icon we can't add it.
You can request it on their github but it's likely in [their roadmap](https://github.com/ryanoasis/nerd-fonts/discussions/716).

lanox:
> Assuming I understood correctly, @lanox This plugin relies on [nerd fonts](https://www.nerdfonts.com/) for it's icons, if they don't provide an icon we can't add it. You can request it on their github but it's likely in [their roadmap](https://github.com/ryanoasis/nerd-fonts/discussions/716).

I think you might be right, looking at the nerdfonts there is nothing for terraform. I might ask the question and see how it goes.

lpuljic:
I think terraform icon is now in nerd-fonts

alex-courtis:
I've had the terraform icon in nerd fonts hack since ~2023 01

PR #185, opened 2023-01-03
Add support for "Containerfile"

adrian5:
Vim recognizes `Containerfile`s as the `dockerfile` filetype. Since they can share the icon with Dockerfiles, I didn't add a separate filetype for them. If that's insufficient, let me know and I'll change it.


alex-courtis (approved):
Thanks for the contribution. Using the same icon does make sense.

It seems that Containerfile has a &filetype of dockerfile so all is well.

PR #186, opened 2023-01-07
feat(default_icon): expose `default_icon` through `get_default_icon`

utilyre:


lua/nvim-web-devicons.lua
```diff
@@ -1704,14 +1704,18 @@ local function setup(opts)
   })
 end
 
+local function get_default_icon(opts)
```
alex-courtis, line 1707: Thank you for raising this one.

Regular use cases following the refactor work just fine.

I did have to set default when calling get_default_icon. Was that your intention? I would expect that calling it with no args would return the default.

```lua
local wdi = require("nvim-web-devicons")

local def = wdi.get_default_icon()
print("opts = {} : " .. vim.inspect(def))

def = wdi.get_default_icon({ default = true })
print("opts = { default = true } : " .. vim.inspect(def))
```

```
opts = {} : nil
opts = { default = true } : {
  color = "#6d8086",
  cterm_color = "66",
  icon = "",
  name = "Default"
}
```

alex-courtis (changes requested):

lua/nvim-web-devicons.lua
utilyre: I thought it'd be a good opportunity to remove duplicate code ([65d7182](https://github.com/nvim-tree/nvim-web-devicons/pull/186/commits/65d71827b418ce2208d622bc529f829d9f1e6bdb)). But now that you mention this odd behavior, I guess just returning the default_icon is good enough.

utilyre: Just removed the opts param.

alex-courtis (approved):

PR #188, opened 2023-01-20
feat: map gql

vivedo:
Map `.gql` to same icon as `.graphql` as it is common to use the alternative filetype

### DIFF
--- a/lua/nvim-web-devicons.lua
+++ b/lua/nvim-web-devicons.lua
@@ -608,6 +608,12 @@ local icons = {
     cterm_color = "199",
     name = "GraphQL"
   },
+  ["gql"] = {
+    icon = "",
+    color = "#e535ab",
+    cterm_color = "199",
+    name = "GraphQL"
+  },
   ["gruntfile"] = {
     icon = "",
     color = "#e37933",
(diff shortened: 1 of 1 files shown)


lua/nvim-web-devicons.lua
```diff
@@ -1482,6 +1488,7 @@ local filetypes = {
   ["go"] = "go",
   ["godot"] = "godot",
   ["graphql"] = "graphql",
+  ["gql"] = "gql",
```
alex-courtis, line 1491: A map of the filetype to the existing icon should be enough `["gql"] = "graphql",`.

Please try that.

alex-courtis (changes requested):

lua/nvim-web-devicons.lua
vivedo: That's exactly what I tried first but I couldn't get it to work

alex-courtis: Right you are; it's necessary when fetching via extension, apologies.

alex-courtis (approved):
Working, many thanks for your contribution!

PR #191, opened 2023-01-24
Add *test.js orange color for react like vs-code

scardenasb:
Add JS icon but in orange for test JS files (eg: App.test.js) like vs-code.

### DIFF
--- a/lua/nvim-web-devicons.lua
+++ b/lua/nvim-web-devicons.lua
@@ -740,6 +740,12 @@ local icons = {
     cterm_color = "185",
     name = "Js",
   },
+  ["test.js"] = {
+    icon = "",
+    color = "#e37933",
+    cterm_color = "173",
+    name = "TestJs",
+  },
   ["json"] = {
     icon = "",
     color = "#cbcb41",


alex-courtis (approved):
Thank you for your contribution.

T-727:
Isn't it sort of odd to merge this considering most languages have `test` extensions yet this only supports Javascript? Some PRs were made to support generic patterns like `*.test.*` but were closed because of potential performance issues. (although none were *actually* measured as far as I'm aware)

alex-courtis:
> Isn't it sort of odd to merge this considering most languages have `test` extensions yet this only supports Javascript? Some PRs were made to support generic patterns like `*.test.*` but were closed because of potential performance issues. (although none were _actually_ measured as far as I'm aware)

These are javascript specific test files, hence they have a javascript specific extension. All the nvim-web-devicons icons are quite targeted to specific file types rather than generic.

T-727:
Hmm, I suppose having specifics is good for differentiating between different languages' tests if they happen to be together.

But wouldn't this mean that if users wanted support for other languages' test icons they would just have to make PRs like this one? Which *could* expand the table quite a lot over time,  resulting in *potentially* worse performance? (just theorizing)

alex-courtis:
> But wouldn't this mean that if users wanted support for other languages' test icons they would just have to make PRs like this one? Which _could_ expand the table quite a lot over time, resulting in _potentially_ worse performance? (just theorizing)

PRs are most gratefully accepted. Lua table lookup by key is cheap, even for large tables.

You can configure your own icons via https://github.com/nvim-tree/nvim-web-devicons#set-an-icon

T-727:
True yes, but in this example the icon is provided by the plugin itself, which encourages users to contribute and get other test icons merged.

PR #195, opened 2023-02-02
fix: vue icon

philopence:


alex-courtis (approved):
Thank you for your contribution

zLinz:
The vue file icon now is question mark?
image

alex-courtis:
Looks similar to the previous in alacritty with Hack Nerd Font Mono



Which font/terminal are you using?

gegoune:
I am getting placeholder icon instead in Wezterm using built in Nerd Font fallback mechanism.

What was reasoning behind that change?

gegoune:
@philopence Does that glyph come from some additional font pack?

alex-courtis:
I'll just revert this for now; we can re-apply once we sort out the root of the issue.

PR #196, opened 2023-02-05
feat: luarocks support

mrcjkb:
nvim-web-devicons is a common dependency across Neovim plugins. Using luarocks may alleviate the need for users to specify their plugins' dependencies in their plugin manager (e.g., vim-plug or packer). See also [this blog post](https://teto.github.io/posts/2021-09-17-neovim-plugin-luarocks.html) for details.

This PR adds a release workflow that publishes this plugin to [LuaRocks](https://luarocks.org/) whenever a tag is pushed,
as well as a rockspec that can be used to release to the `dev` channel. 

For the release workflow to work, someone with a LuaRocks account will have to add their API key to this repo's secrets.

Note that I have added a shield to the readme that assumes the existence of an `nvim-tree` LuaRocks account.

github-add-luarocks-api-key


.github/workflows/release.yml
```diff
@@ -0,0 +1,23 @@
+name: "release"
+on:
+  push:
+    tags:
+      - 'v*'
```
alex-courtis, line 5: What's the convention for versions here? SEMVER e.g. `v1.2.3`?

It might be a good time for us to decide on an nvim-tree version scheme @gegoune

.github/workflows/release.yml
```diff
@@ -0,0 +1,23 @@
+name: "release"
+on:
+  push:
+    tags:
```
alex-courtis, line 4: Can we use a branch? Users of some plugin managers prefer branches https://github.com/nvim-tree/nvim-tree.lua/issues/1944

alex-courtis (commented):
This is fantastic! Thank you @mrcjkb

.github/workflows/release.yml
mrcjkb: I'm not sure if it's a [hard requirement](https://github.com/luarocks/luarocks/wiki/Rockspec-format), but the convention for luarocks is semver.

.github/workflows/release.yml
mrcjkb: It should be possible to trigger the build only for certain branches.
But at the moment, the GitHub action uses tags to generate the release versions.

I do plan to see if it's possible to use GitHub releases, too: https://github.com/nvim-neorocks/luarocks-tag-release/issues/8

alex-courtis: > But at the moment, the GitHub action uses tags to generate the release versions.

Good enough for me. tags _are_ the correct approach for versions.

If there is user pushback they can investigate options.

.github/workflows/release.yml
alex-courtis: Thank you.

gegoune: vMajor.Minor.Patch seems to be standard. Lets use that.

.github/workflows/release.yml
```diff
@@ -0,0 +1,23 @@
...
+jobs:
+  luarocks-upload:
+    runs-on: ubuntu-22.04
+    steps:
+      - uses: actions/checkout@v3
+      - name: LuaRocks Upload
+        uses: nvim-neorocks/luarocks-tag-release@v1.0.2
+        env:
+          LUAROCKS_API_KEY: ${{ secrets.LUAROCKS_API_KEY }}
+        with:
```
gegoune, line 15: Seems like this is already part of rockspec. Does it need to be repeated here as well?

mrcjkb: The scm-1.rockspec isn't part of the workflow (and isn't necessary for it).
I added it because it can be used for local installs or to publish a `dev` version, which tells luarocks to build the package using the latest git source (see the `source.url`).

I guess it's more useful for plugins that have dependencies.

mrcjkb:
@alex-courtis @gegoune I've bumped the version of the action (it now auto-detects plugin directories) and removed the rockspec, since it's a redundant second source of truth.

What's the status on this from your side?

gegoune:
I guess missing bit still is

> For the release workflow to work, someone with a LuaRocks account will have to add their API key to this repo's secrets.

@alex-courtis Have you got an account there?

alex-courtis:
> I guess missing bit still is
> 
> > For the release workflow to work, someone with a LuaRocks account will have to add their API key to this repo's secrets.
> 
> @alex-courtis Have you got an account there?

This rock is managed by an organisation of some sort: https://luarocks.org/modules/openresty/lua-cjsone

It's the same (except case) name of the github organisation.

We can't log in to github as an organisation and I can't register github at luarocks as anything but myself.

Email is required for non-github registration.

Possibilities:
1. create an nvim-tree github user
1. create a luarocks user
1. use a personal github account

1 seems manageable, although we'd need a few email addresses for password resets etc.

mrcjkb:
@alex-courtis I don't think it's possible to create a github user of the same name as an existing organisation.

@teto how did you do this with the neovim and nvim-treesitter accounts on LuaRocks? I guess it was a non-github registration?

teto:
so IIRC I've created a "neovim" user and then I've generated several LUAROCKS_API_KEY, one for neovim and one for nvim-treesitter

alex-courtis:
> so IIRC I've created a "neovim" user and then I've generated several LUAROCKS_API_KEY, one for neovim and one for nvim-treesitter

That sounds reasonable. Is this the neovim user? https://luarocks.org/modules/neovim

How do you manage the user and its API keys? How are that user's credentials maintained? Considering a  "5 years time" situation when, say, a new key is needed.

teto:
the neovim user is meant to provide API keys for "official" packages, ie., maintained by the neovim core team (or close, like nvim-treesitter's team). In your case you would need to register yourself on luarocks.org (github account works) and create the LUAROCKS_API_KEY yourself.

alex-courtis:
This is not practical until we get a deployment strategy that is not tied to an individual.

Please reopen if you have any further ideas.

mrcjkb:
@alex-courtis So creating a `nvim-tree` luarocks.org account that is not linked to a GitHub account would not be an option for you?

Asking, because we have been working on [rocks.nvim](https://github.com/nvim-neorocks/rocks.nvim), which facilitates using luarocks to manage plugins.

alex-courtis:
> @alex-courtis So creating a `nvim-tree` luarocks.org account that is not linked to a GitHub account would not be an option for you?
> 
> Asking, because we have been working on [rocks.nvim](https://github.com/nvim-neorocks/rocks.nvim), which facilitates using luarocks to manage plugins.

rocks.nvim looks incredible! Thank you for all your work.

A username/password account does indeed allow an API key.

Unfortunately we still have the problem of authentication in the far future: if the password is lost or the API key expires we cannot recover it via, say, email.

Is there prior art other nvim itself and nvim-treesitter? Is there another plugin we could emulate?

alex-courtis:
Pragmatic but fragile solution:
* create a luarocks account for nvim-tree organisation with a username/password
* an email must be provided, `none@example.com` works
* record the credentials in a secure location inside the organisation
* retrieve / generate API keys as needed via luarocks.org

In the event of credential loss or expiration luarocks.org would need to manually recreate or reactivate the account based on the claim of an organisation member.

@gegoune @Akmadan23 what are your thoughts?

Akmadan23:
> Pragmatic but fragile solution:
> 
>   * create a luarocks account for nvim-tree organisation with a username/password
>   * an email must be provided, `none@example.com` works
>   * record the credentials in a secure location inside the organisation
>   * retrieve / generate API keys as needed via luarocks.org

Seems a good idea... We could create an email account for the nvim-tree organization so that we can all access any important email from luarocks and even recover/recreate the password if we lost it.

alex-courtis:
An email and luarocks.org account nvim-tree has been created.

All details recorded in the private nvim-tree repository, including recoveries etc.

alex-courtis:
Done: manual run for v0.99 tag

https://luarocks.org/modules/nvim-tree/nvim-web-devicons

Unfortunately the label was not set; I've done it manually.

mrcjkb:
> Done: pushed changes to master and manually ran for v0.99 tag

🚀🌘
Thanks for all the effort!

> Unfortunately the label was not set; I've done it manually.

Yes, that's a shortcoming that needs to be fixed.

> Is there anything else we need to do here

Looks all good to me.
I'll close this PR 😃

PR #197, opened 2023-02-07
feat: Add `strict` option

### LINKED ISSUE #194: Feature proposal (ready to do a PR): distinguish between filename-based icons and file extension-based icons
So I've been using NvimTree (which as you know uses nvim-web-devicons) for the past few weeks, and noticed something that seemed not very "as-intended" to me.
So as far as I understood by looking at implementation basically nvim-web-devicons has one table and both icons for filenames and icons for file extensions reside there.
This results in various icons being displayed for files which happen to have a name that matches the extension in this table, example:

Screen Shot 2023-01-31 at 8 27 20 AM
---

I've played around with code a bit and came up with completely backward-compatible solution that fixed that problem; example:

Screen Shot 2023-01-31 at 8 28 32 AM

As you can see now files that do not have extension have a default icon, as I would personally expect.
Basically here's what I changed:
* there are now three icon tables: `icons_by_filename`, `icons_by_file_extension` and `icons`
* there is a new option `strict` (didn't came up with a better name yet) which is false by default but when true - `get_icon` looks up `icons_by_filename` for icon by `name`, and `icons_by_file_extension` for icon by `ext`
* icons by filename and extension are provided by user with `override.by_filename` and `override.by_extension`

Icons from both tables are still "merged" into `icons` table and user can still provide icons for both filenames and extensions in `override`, so this feature is completely backward compatible.

What do you think? Should I do a PR or should I just keep my fork and use it?

carljohnson93:
This option allows to strictly select icon by filename or by extension, to prevent cases when filename happens to match some extension.
Closes #194

### DIFF
--- a/lua/nvim-web-devicons.lua
+++ b/lua/nvim-web-devicons.lua
@@ -1746,8 +1746,8 @@ local function setup(opts)
     default_icon = user_icons.override.default_icon
   end
 
-  local user_filename_icons = user_icons.override and user_icons.override.by_filename
-  local user_file_ext_icons = user_icons.override and user_icons.override.by_extension
+  local user_filename_icons = user_icons.override_by_filename
+  local user_file_ext_icons = user_icons.override_by_extension
 
   icons = vim.tbl_extend(
     "force",
@@ -1759,11 +1759,9 @@ local function setup(opts)
 
   if user_filename_icons then
     icons_by_filename = vim.tbl_extend("force", icons_by_filename, user_filename_icons)
-    icons.by_filename = nil
   end
   if user_file_ext_icons then
     icons_by_file_extension = vim.tbl_extend("force", icons_by_file_extension, user_file_ext_icons)
-    icons.by_extension = nil
   end
 
   table.insert(icons, default_icon)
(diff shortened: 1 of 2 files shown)


lua/nvim-web-devicons.lua
```diff
@@ -1712,13 +1736,35 @@ local function setup(opts)
...
+  end
+
   global_opts.color_icons = if_nil(user_icons.color_icons, global_opts.color_icons)
 
   if user_icons.override and user_icons.override.default_icon then
     default_icon = user_icons.override.default_icon
   end
 
-  icons = vim.tbl_extend("force", icons, user_icons.override or {})
+  local user_filename_icons = user_icons.override and user_icons.override.by_filename
```
alex-courtis, line 1749: Let's change the setup schema to
```lua
{
  override = {
    ---
  },
  override_by_filename = {
    ---
  },
  override_by_extension = {
    ---
  },
---
}
```

Having the specific overrides next to the icons is quite confusing.

alex-courtis (changes requested):
Looking good!

Tested:

package.json
bleh.json
bleh.package.json

strict: nil, false, true

get_icon
get_icon_color_by_filetype
get_icon_colors
setup

Did not test setup override; we need to refactor first.

Please:

- [x] simplify setup schema
- [ ] add new setup options to README

carljohnson93:
Changed overrides setup schema. I'm not very good at writing docs, will try my best

alex-courtis (approved):
Tested OK, readme looks good.

PR #199, opened 2023-02-11
fix: support ext convention

nobuhikosawai:
I saw this https://github.com/nvim-tree/nvim-web-devicons/pull/191 and I like the idea of supporting `*.test.js`, `*.stories.tsx`, etc because these are quite common conventions.

The problem is currently this is not working as intended. (Only `test.js` highlights but not `foo.test.js`)

image

This PR is to fix this issue.
image

*Note*
Whether a plugin can reflect this change depends on the plugin's implementation.
(I see some attempt to support option to switch the way to get icons like this PR: https://github.com/akinsho/bufferline.nvim/pull/669)

image

### DIFF
--- a/lua/nvim-web-devicons.lua
+++ b/lua/nvim-web-devicons.lua
@@ -1736,7 +1736,7 @@ local function get_default_icon()
 end
 
 local function get_icon(name, ext, opts)
-  ext = ext or name:match("^.*%.(.*)$") or ""
+  ext = ext or name:match("^.-%.(.*%.*.*)$") or ""
   if not loaded then
     setup()
   end
@@ -1759,7 +1759,7 @@ local function get_icon_by_filetype(ft, opts)
 end
 
 local function get_icon_colors(name, ext, opts)
-  ext = ext or name:match("^.*%.(.*)$") or ""
+  ext = ext or name:match("^.-%.(.*%.*.*)$") or ""
   if not loaded then
     setup()
   end


lua/nvim-web-devicons.lua
```diff
@@ -1736,7 +1736,7 @@ local function get_default_icon()
 end
 
 local function get_icon(name, ext, opts)
-  ext = ext or name:match("^.*%.(.*)$") or ""
+  ext = ext or name:match("^.*%.(.*%..*)$") or name:match("^.*%.(.*)$") or ""
```
alex-courtis, line 1739: Could we simply match `"^.*%.(.*%..*)$"` without `"^.*%.(.*)$"`? Greedy match should cover all existing cases...

alex-courtis (changes requested):
Your timing is excellent; I've just been testing #197

This one's a bug, not a feature and doesn't break anything.

Tested OK for cases you referenced, as well as regular cases.

We'll merge that one first.

lua/nvim-web-devicons.lua
nobuhikosawai: @alex-courtis Thanks for your feedback. I've updated the regexp pattern to match both 'foo.test.js' and 'foo.js'. Could you double check?

alex-courtis (approved):
Tested OK:

bleh.stories.tsx
bleh.tsx
stories.tsx

Will merge #197 first

alex-courtis (approved):
Merged master, retested OK.

hthuong09:
But then again it breaks other types.

For example, I have a file called `cat.service.ts`, the extension is now `service.ts` and it does not match any icon.

I would suggest adding another variable to this case and checking it first before the actual extension.

alex-courtis:
Reported #205

Reverting.

PR #208, opened 2023-02-19
feat: more accurate cterm colors

mg979:
Recalculated with the algorithm from lifepillar/vim-colortemplate, that gives better approximations.

### DIFF
--- a/lua/nvim-web-devicons.lua
+++ b/lua/nvim-web-devicons.lua
@@ -53,55 +53,55 @@ local icons_by_filename = {
   [".DS_Store"] = {
     icon = "",
     color = "#41535b",
-    cterm_color = "59",
+    cterm_color = "239",
     name = "DsStore",
   },
   [".gitattributes"] = {
     icon = "",
     color = "#41535b",
-    cterm_color = "59",
+    cterm_color = "239",
     name = "GitAttributes",
   },
   [".gitconfig"] = {
     icon = "",
     color = "#41535b",
-    cterm_color = "59",
+    cterm_color = "239",
     name = "GitConfig",
   },
   [".gitignore"] = {
     icon = "",
     color = "#41535b",
-    cterm_color = "59",
+    cterm_color = "239",
     name = "GitIgnore",
   },
   [".gitlab-ci.yml"] = {
     icon = "",
     color = "#e24329",
-    cterm_color = "166",
+    cterm_color = "196",
     name = "GitlabCI",
   },
   [".gitmodules"] = {
     icon = "",
     color = "#41535b",
-    cterm_color = "59",
+    cterm_color = "239",
     name = "GitModules",
   },
   [".gvimrc"] = {
     icon = "",
     color = "#019833",
-    cterm_color = "29",
+    cterm_color = "28",
     name = "Gvimrc",
   },
   [".npmignore"] = {
     icon = "",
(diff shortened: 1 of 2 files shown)


alex-courtis (changes requested):
This is fantastic! Thank you.

Please:

- [ ] add to scripts so that we may do this again in the future

mg979:
Added.

scripts/cterm-colors-generator.lua
```diff
@@ -0,0 +1,43 @@
...
+-- it's best to have the main file already open in another window
+
+-- IMPORTANT: the `cterm_color` key must always be below the `color` key of the
+--            same icon. Currently they are.
+
+-- :source this file when ready
+
+local fn = vim.fn
+
+vim.cmd("packadd vim-colortemplate")
```
alex-courtis, line 12: Works nicely, not sure we need this, as it will be automatically loaded from &rtp

scripts/cterm-colors-generator.lua
```diff
@@ -0,0 +1,43 @@
+-- vim-colortemplate is needed
```
alex-courtis, line 1: How about:

```
-- Plugin lifepillar/vim-colortemplate must be available on &runtimepath
--
```

alex-courtis (changes requested):
Nice work

scripts/cterm-colors-generator.lua
mg979: If you're sure it's not needed I'll remove it.

scripts/cterm-colors-generator.lua
mg979: Sure.

mg979:
Updated.

alex-courtis (approved):
Many thanks for your contribution

lua/nvim-web-devicons.lua
```diff
@@ -1492,13 +1492,13 @@ local icons_by_file_extension = {
...
     color = "#519aba",
-    cterm_color = "67",
+    cterm_color = "74",
     name = "Solidity",
   },
   ["prisma"] = {
     icon = "卑",
     color = "#ffffff",
-    cterm_color = "white",
+    cterm_color = "231",
```
alex-courtis, line 1501: big win

scripts/cterm-colors-generator.lua
mg979: As long as it's in `start` it's not needed.

lua/nvim-web-devicons.lua
mg979: It's the same color.

The script only checks for numbers actually, not for color names. Originally I used a macro, that plugin also works with color names. Maybe I should update the script so that it also catches color names.

mg979: What would you do instead?

alex-courtis: It's correct... I'm just happy we are using colours consistently now ;)

mg979: But it doesn't really matter, as long as gui colors are written in html format, even a zero will be turned into the right color. It's not a big issue.

mg979: That's ok then 👍🏻

PR #215, opened 2023-02-28
feat: js/ts/jsx/tsx consistent test/spec icons

ptakpatryk:
I like that someone has finally added test.ts files, but I think we're missing some consistency now especially for react enthusiasts ;)
So here I come with the setup that I had on my config files (this is what VS Code users would normally get by default).

Before:
Screenshot 2023-02-28 at 20 16 19

After:
Screenshot 2023-02-28 at 20 18 48


alex-courtis (approved):
This is great, thank you.

Updated cterm_color

Issue #206, opened 2023-02-19
Compound Extensions


alex-courtis:
See #191, rolled back.

Only `test.js` highlights but not `foo.test.js`

/cc [nobuhikosawai](https://github.com/nobuhikosawai)

gegoune:
How expensive/inefficient iterating over each `.` segment and comparing against available icons would be?

alex-courtis:
> How expensive/inefficient iterating over each . segment and comparing against available icons would be?

That sounds the simplest. More efficient than string.find etc.

nobuhikosawai:
I'll try to work on this again.

PR #216, opened 2023-03-02
fix: handle extension with multiple dots

nobuhikosawai:
fixes: https://github.com/nvim-tree/nvim-web-devicons/issues/206

I took @gegoune's approach on this comment (https://github.com/nvim-tree/nvim-web-devicons/issues/206#issuecomment-1436045476)

Before
Screenshot 2023-03-03 at 0 56 39

After
Screenshot 2023-03-03 at 0 57 32

(Note that I have set the custom icon on my config for `.stories.tsx` )

### DIFF
--- a/lua/nvim-web-devicons.lua
+++ b/lua/nvim-web-devicons.lua
@@ -1868,6 +1868,21 @@ local function get_default_icon()
   return default_icon
 end
 
+-- recursively iterate over each segment separated by '.' to parse extension with multiple dots in filename
+local function iterate_multi_dotted_extension(name, icon_table)
+  if name == nil then
+    return nil
+  end
+
+  local compound_ext = name:match("%.(.*)")
+  local icon = icon_table[compound_ext]
+  if icon then
+    return icon
+  end
+
+  return iterate_multi_dotted_extension(compound_ext, icon_table)
+end
+
 local function get_icon_by_extension(name, ext, opts)
   local is_strict = if_nil(opts and opts.strict, global_opts.strict)
   local icon_table = is_strict and icons_by_file_extension or icons
@@ -1876,18 +1891,7 @@ local function get_icon_by_extension(name, ext, opts)
     return icon_table[ext]
   end
 
-  if name == nil then
-    return nil
-  end
-
-  -- recursively iterate over each segment separated by '.' to parse extension with multiple dots in filename
-  local compound_ext = name:match("%.(.*)")
-  local icon = icon_table[compound_ext]
-  if icon then
-    return icon
-  end
-
-  return get_icon_by_extension(compound_ext, ext, opts)
+  return iterate_multi_dotted_extension(name, icon_table)
 end
 
 local function get_icon(name, ext, opts)
--- a/lua/nvim-web-devicons.lua
+++ b/lua/nvim-web-devicons.lua
@@ -1868,10 +1868,31 @@ local function get_default_icon()
   return default_icon
 end
 
+local function get_icon_by_extension(name, ext, opts)
+  local is_strict = if_nil(opts and opts.strict, global_opts.strict)
+  local icon_table = is_strict and icons_by_file_extension or icons
+
+  if ext ~= nil then
+    return icon_table[ext]
+  end
+
+  if name == nil then
+    return nil
+  end
+
+  -- recursively iterate over each segment separated by '.' to parse extension with multiple dots in filename
+  local compound_ext = name:match("%.(.*)")
+  local icon = icon_table[compound_ext]
+  if icon then
+    return icon
+  end
+
+  return get_icon_by_extension(compound_ext, ext, opts)
+end
+
 local function get_icon(name, ext, opts)
 	name = name:lower()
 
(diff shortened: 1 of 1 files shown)


lua/nvim-web-devicons.lua
```diff
@@ -1857,8 +1857,29 @@ local function get_default_icon()
   return default_icon
 end
 
+local function get_icon_by_extension(name, ext, opts)
```
alex-courtis, line 1860: I _think_ this can be added to the API.

Please do so if appropriate, updating the readme.

alex-courtis (changes requested):

Please:

- [ ] extract API if appropriate
- [ ] resolve merge conflicts
- [x] extract a small recursive function

lua/nvim-web-devicons.lua
nobuhikosawai: @alex-courtis This one, I am actually not very sure. `get_icon_by_extension` will ignore `icon_by_filename` so filename like `docker-compose.yaml` will be treated as just `yaml`.
I am not sure how much useful this is.

Still if you still think this is appropriate to add to the API, then I'll update it.

nobuhikosawai:
I've done these 

- [ ]  extract API if appropriate
- [x]  resolve merge conflicts
- [x]  extract a small recursive function

lua/nvim-web-devicons.lua
alex-courtis: That's fair enough. It doesn't really add anything but confusion.

The users can achieve everything via current API.

alex-courtis:
Looks good; I just cannot remember the original problem. I can't replicate the before/after in the description.

What's the test case? I'm trying `icons_by_file_extension`
```lua
  ["stories.tsx"] = {
    icon = "X",
    color = "#519aba",
    cterm_color = "74",
    name = "StoriesTsx",
  },
``

**Comment by nobuhikosawai** [comment]
### Steps to reproduce

This is the minimum steps to reproduce the original problem:
1. Set `strict=true` on nvim-web-devicon.
```
require'nvim-web-devicons'.setup {
 strict = true,
}
```
2. Install [telescope.nvim](https://github.com/nvim-telescope/telescope.nvim), one of the most popular fuzzy finder plugin. This plugin supports integration with nvim-web-devicons.
3. Add `foo.test.js` in your project. (`stories.tsx` has nothing to do with reproducing the problem. Apologies for the confusion.)
4. Search "test.js" by using `:Telescope find_files`.

**Expected behavior:**
Since "test.js" is in the `icons_by_file_extension` table, `foo.test.js` should be displayed with the test.js icon.
image

**Actual behavior:**
`foo.test.js` appears with `js` icon, not `test.js` icon.
image

You can reproduce this behavior with many other plugins (such as [lualine.nvim](https://github.com/nvim-lualine/lualine.nvim)) that support integration with nvim-web-devicons.

Please note that this is not a issue on nvim-tree because it has already been solved within the plugin itself after this PR (https://github.com/nvim-tree/nvim-tree.lua/pull/423).

### My test cases
As I have written as an example in my description, I have tested the following filenames:
- foo.ts => Showing ts icon.
- foo.rc.lua => Showing lua icon since there is no "rc.lua" in the icon table. (This was raised in https://github.com/nvim-tree/nvim-web-devicons/issues/205 with my original implementation https://github.com/nvim-tree/nvim-web-devicons/pull/199)
- foo.stories.ts => Tested if the user configured icons are resolved correctly. This also works fine.
- foo.some.long.compound.ext.py => This filename has more than 2 dots, but it resolves to the `py` icon correctly.

(This is off-topic, but maybe we should have a test code.)

### Relevant issue not mentioned above.
https://github.com/nvim-tree/nvim-web-devicons/issues/36

**Comment by alex-courtis** [comment]
> 1. Set `strict=true` on nvim-web-devicon.

Thank you; that seems to be the bit I was missing.

> foo.test.js appears with js icon, not test.js icon.

I cannot replicate this on master. I can replicate it on your fork's master c2c2317 however the test.js and js icons are the same; test.js was updated at 87c3ded

I think I understand now:
```lua
local result = ""

local function tst(name, ext)
	local icon, hl = require('nvim-web-devicons').get_icon(name, ext, { strict = true })
	result = string.format("%s, %s (%s) %s %s", result, name, ext, icon, hl)
end

tst("test.js", "js")
tst("test.js", nil)

tst("foo.test.js", "test.js")
tst("foo.test.js", "js")
tst("foo.test.js", nil)

print(result)
```

master:
`test.js (js)  DevIconJs, test.js (nil)  DevIconJs, foo.test.js (test.js)  DevIconTestJs, foo.test.js (js)  DevIconJs, foo.test.js (nil)  DevIconJs`

branch:
`test.js (js)  DevIconJs, test.js (nil)  DevIconJs, foo.test.js (test.js)  DevIconTestJs, foo.test.js (js)  DevIconJs, foo.test.js (nil)  DevIconTestJs`

alex-courtis (approved):
Many thanks for your patience.

alex-courtis:
> (This is off-topic, but maybe we should have a test code.)

Yes, that would greatly improve quality and reduce regressions.

I'd be most grateful if you could put something together!

PR #217, opened 2023-03-03
Normalize case

mskelton:
Fixes https://github.com/nvim-tree/nvim-web-devicons/issues/44

This updates all the icon mappings to be fully lower case and then sanitizes the name passed in to `get_icon` to be lowercase as well. This makes the plugin now case insensitive to better support variations in filename casing.


alex-courtis (approved):
Thank you so much. GNUmakefile was not pretty.

PR #218, opened 2023-03-06
ci: add stylua, luacheck, luarc, editorconfig, pre-commit hooks

alex-courtis:
Mostly sourced from https://github.com/nvim-tree/nvim-tree.lua


scripts/setup-hooks.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/sh
+
+ln -sf ../../.hooks/pre-commit.sh .git/hooks/pre-commit
```
mg979, line 3: Is this path right? From where is this run?

.hooks/pre-commit.sh
```diff
@@ -0,0 +1,4 @@
+#!/usr/bin/env bash
```
gegoune, line 1: Why that script and not regular pre-commit configuration file?

CONTRIBUTING.md
```diff
@@ -48,7 +48,24 @@ Please commit both `nvim-web-devicons.lua` and `nvim-web-devicons-light.lua`
 
 ## Styling And Format
 
-TODO @alex-courtis
+Code is formatted using luacheck and linted using stylua.
```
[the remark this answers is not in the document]
alex-courtis: Good catch!

scripts/setup-hooks.sh
alex-courtis: https://github.com/nvim-tree/nvim-web-devicons/pull/218/files#diff-eca12c0a30e25b4b46522ebf89465a03ba72a03f540796c979137931d8f92055R65

However... I think there is a better method https://github.com/nvim-tree/nvim-web-devicons/pull/218#discussion_r1126748422

mg979:
Can't we just avoid git hooks and have a shell script that runs luacheck/stylua/make? You'll have to do `make` anyway, you can't automate everything.

.hooks/pre-commit.sh
alex-courtis: Does that mean the user needs https://pre-commit.com/ ?

I've had some... trouble... convincing fellow developers to use pre-commit mechanisms.

We do, however, need the same style/lint CI as nvim-tree. Future PR...

alex-courtis:
> Can't we just avoid git hooks and have a shell script that runs luacheck/stylua/make? You'll have to do `make` anyway, you can't automate everything.

That is sensible. It sidesteps the git hooks issue.

Added to makefile and updated doc.

mg979:
Much better thanks :)

.hooks/pre-commit.sh
gegoune: If contributors don't have pre-commit set up it will be a noop, so no real downside I think. We can then use pre-commit on CI to run the same checks that local pre-commit would do and fail if any files would be changed by set up tasks, which would then fallback to contributor to fix those issues.

alex-courtis: > If contributors don't have pre-commit set up it will be a noop

Sounds good to me. It's really up to them to set things up locally.

> CI to run the same checks that local pre-commit would do

That works beautifully for nvim-tree.

Can I please ask you to make the changes here? We can set up CI as soon as this is merged.

gegoune: @alex-courtis please see #233

alex-courtis: Many thanks!

PR #238, opened 2023-03-27
ci: add linter on PR

gegoune:


alex-courtis:
I think it's OK to just merge these changes to master; no changes to "release code" is being made.

if there's an issue with a PR we can address it then.

PR #239, opened 2023-03-28
dummy test 238 valid changes

alex-courtis:

PR #240, opened 2023-03-28
dummy test 238 bad style

alex-courtis:


alex-courtis:
CI [failed as expected](https://github.com/nvim-tree/nvim-web-devicons/actions/runs/4537873645) for the first commit. This is just fantastic!

As per #239 it didn't run for subsequent runs.

PR #243, opened 2023-03-28
ci: ensure PR subjects follow semantic commit spec

gegoune:


alex-courtis (approved):
I never even knew this was an official spec... many thanks!

PR #246, opened 2023-04-02
feat: use official icon for `nim`

tobealive:
Nim's offical icon is supported meanwhile. I thought we could use it.

Screenshot_20230402_192854
Screenshot_20230402_192935


alex-courtis (approved):

PR #247, opened 2023-04-02
feat: add icon for `V`

tobealive:
This PR adds an icon for the `V` language.

Just `"v"` was updated from currently linking to verilog, since it used by treesitter and lsp for the V language. `"vlang"` was added which is also used by those two, and many other `V` plugins.

### DIFF
--- a/lua/nvim-web-devicons-light.lua
+++ b/lua/nvim-web-devicons-light.lua
@@ -1397,10 +1397,10 @@ local icons_by_file_extension = {
     name = "Txt",
   },
   ["v"] = {
-    icon = "",
-    color = "#017226",
-    cterm_color = "22",
-    name = "Verilog",
+    icon = "𝗩",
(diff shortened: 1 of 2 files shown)


alex-courtis:
Unfortunately `.v` is the verlog file extension; this change would break verilog users.

verilog has been around for many years whilst the V programming language is very new, at 0.3.

Perhaps we could use `.V` for the V programming language.

tobealive:
Thanks tackling this @alex-courtis!

I

tobealive:
Slipped out clicking close and then my internet went down. Maximum fail.

Updated the half finished comment @alex-courtis

alex-courtis:
> Maybe this is an argument for the PR, else it looks like it's an actual real conflict and a user based override might be the only solution.

Unfortunately it's all we can do.

The V language developers should have thought more carefully ;)

PR #250, opened 2023-04-07
Update README: Fix signature of `set_default_icon`

bew:
Ref:
https://github.com/nvim-tree/nvim-web-devicons/blob/0568104bf8d0c3ab16395433fcc5c1638efc25d4/lua/nvim-web-devicons.lua#L2006


gegoune:
Thank you!

PR #264, opened 2023-05-01
fix: update to Nerd Fonts 3 code points

vlada-dudr:
Nerd Fonts 3 changed code points of some icons, this updates it according  to table [here](https://github.com/ryanoasis/nerd-fonts/issues/1059)

This will break everyone being on Nerd 2, but without it everyone on version 3 is broken also...

fixes #265

PR #266, opened 2023-05-04
fix: fixed icon of vue

jfbeatit:
Fix reason: I use the font `SauceCodePro Nerd Font Mono `, after i update this plugin to the latest version, the vue icon can not display anymore.

### DIFF
--- a/lua/nvim-web-devicons.lua
+++ b/lua/nvim-web-devicons.lua
@@ -1453,7 +1453,7 @@ local icons_by_file_extension = {
     name = "Vim",
   },
   ["vue"] = {
-    icon = "",
+    icon = "﵂",
     color = "#8dc149",
     cterm_color = "113",
     name = "Vue",


gegoune:
Try updading your font instead, this icon has been recently changed to more accurate one.

alex-courtis:
Let's merge #264 first

gegoune:
Not sure we want to merge this one, Alex. Vue icon has been very recently updated to one that's reverted here.

If we are using proper icons that are only available in recent font versions we will endlessly get requests like this while solution is to update fonts. 

We will shortly provide compat tag for version 2 icons. That might solve the issue this pr tries to address.

Or we should come up with proper policy on how we handle those things.

lua/nvim-web-devicons.lua
```diff
@@ -1453,7 +1453,7 @@ local icons_by_file_extension = {
     name = "Vim",
   },
   ["vue"] = {
-    icon = "",
+    icon = "﵂",
```
alex-courtis, line 1456: Unfortunately this is a nerd font v2 icon; we need a v3 icon.

See #264

You'll need to run nerdfix.

alex-courtis (changes requested):

alex-courtis:
> Not sure we want to merge this one, Alex. Vue icon has been very recently updated to one that's reverted here.
> 
> If we are using proper icons that are only available in recent font versions we will endlessly get requests like this while solution is to update fonts.

It looks like the vue icon on master is the correct v3 icon.

> We will shortly provide compat tag for version 2 icons. That might solve the issue this pr tries to address.
> 
> Or we should come up with proper policy on how we handle those things.

The tag should be enough.

I have Hack Nerd Font Mono set as my browser font, so I can quickly see in the PR whether the icon is a valid (v3) icon.

alex-courtis:
> Or we should come up with proper policy on how we handle those things.

We _could_ maintain a v2 branch rather than a tag, for reverting v3->v2 icons.

That's a lot of work and the branches will completely diverge.

alex-courtis:
Optimistically closing: @jfbeatit can you please update your font to v3 and test again.

We'll reopen if that doesn't work.

PR #274, opened 2023-05-19
Update nvim-web-devicons-light.lua

WuKuohaoMaster:


alex-courtis:
This should be generated and submitted as part of the same PR. See https://github.com/nvim-tree/nvim-web-devicons/blob/master/CONTRIBUTING.md

PR #277, opened 2023-05-24
feat: add mapping for unlicense

snelling-a:


alex-courtis (approved):

PR #285, opened 2023-06-23
fix: correct zig icon f0e7->e6a9

bauripalash:
The previous icon was just a thunder. Added correct icon


alex-courtis (approved):
Looks great, many thanks for your contribution.

Issue #168, opened 2022-10-17
Add icon for Dockerfile or dockerfile with extensions


olexsmir:
Example:
![image](https://user-images.githubusercontent.com/50584123/196205115-3c8cab4a-b781-4ee7-acce-01a8b5471e15.png)

Flo-Slv:
Hi guys, any news on this one ?!

alex-courtis:
[Pull requests](https://github.com/nvim-tree/nvim-web-devicons/blob/master/CONTRIBUTING.md) are gratefully appreciated.

BBboy01:
This issue has been resolved by #322, but not suitable for `Dockerfile.xxx` some kind of "prefixed" file type, is this ok to add a `icons_by_file_prefix` table?

alex-courtis:
I hadn't noticed this one, thanks for bringing it up.

Yes, we could add an `icons_by_file_prefix`, with name and extension taking precedence so as not to break existing behaviour.

PR #289, opened 2023-07-27
feat(#168): add Dockerfile icon

wael-jaber:


alex-courtis (changes requested):
The `dockerfile` filetype is already specified and a `dockerfile` file name is specified, which is compared case insensitively.

Both are currently working - what's the issue you are experiencing here?

wael-jaber:
@alex-courtis  hello , sorry I seen the pr and I just assumed it isn't there but now I see what is my problem 
image
it isn't showing because it has an extension format (.prod) in my case. then maybe this will be a feature request to have last fall back to compare to substring (starting from the start of the string) however that can cause some side effect. wdyt ?

alex-courtis:
The filetype may be set for the file via a `:help modeline` e.g.
`# vim:ft=dockerfile`

web-devicons can fetch the icon for that filetype.

PR #290, opened 2023-08-05
feat: add vlang .vsh

z7ealth:
Added Vlang Icon support.

### DIFF
--- a/lua/nvim-web-devicons-light.lua
+++ b/lua/nvim-web-devicons-light.lua
@@ -1444,12 +1444,6 @@ local icons_by_file_extension = {
     cterm_color = "22",
     name = "Txt",
   },
-  ["v"] = {
-    icon = "",
-    color = "#556d8c",
-    cterm_color = "24",
-    name = "Vlang",
-  },
   ["vsh"] = {
     icon = "",
     color = "#556d8c",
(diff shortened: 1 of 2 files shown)


gegoune:
Please also add light colour version.

z7ealth:
HI @gegoune,

I added both the light theme icon, and the icon for vlang's shell files (.vsh)

gegoune:
Thanks. Seems like icon for `v` already exists.

z7ealth:
@gegoune there was one but for Verilog, is there a way to use the right icon when the file extension is the same in two languages?
In this case Vlang and Verilog.

alex-courtis:
We did previously discuss this: #247

Verilog existed long before v and we don't want to break verilog users.

> there was one but for Verilog, is there a way to use the right icon when the file extension is the same in two languages?

I think the best course of action is for you to override: https://github.com/nvim-tree/nvim-web-devicons#set-an-icon

I'm happy to take the vsh icon.

z7ealth:
@alex-courtis okok, i removed the v option from vlang, only vsh will be added, for .v files, i'll use the override option then, greetings.

alex-courtis (approved):
Used generated cterm and light colours.

PR #293, opened 2023-08-15
feat: svelte config icon

TheBlckbird:
This uses the svelte icon for the `svelte.config.js` file.
Many other editors assign specific icons to configs of a framework and I think this should be done here as well.

### DIFF
--- a/lua/nvim-web-devicons-light.lua
+++ b/lua/nvim-web-devicons-light.lua
@@ -83,6 +83,12 @@ local icons_by_filename = {
     cterm_color = "91",
     name = "SettingsJson",
   },
+  ["svelte.config.js"] = {
+    icon = "",
+    color = "#bf2e00",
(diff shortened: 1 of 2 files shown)


gegoune:
Thanks!

TheBlckbird:
Old icon:
Bildschirmfoto 2023-08-16 um 23 45 51
New icon:
Bildschirmfoto 2023-08-16 um 23 46 10

I've just noticed that the old and new icons are shown with different colors although they both are assigned the same color.

I'm using [NvChad](https://nvchad.com/) with the `rosepine` theme.
This happens with other themes as well.

alex-courtis:
I'm seeing this:


Do you have some nvim-web-devicons configuration that is overriding / adding?

TheBlckbird:
@alex-courtis I'm using NvChad, so that might be the problem

PR #302, opened 2023-09-07
feat: add `.cts` and `.mts` icons

auvred:
I copied the icon that is used for the `.ts` extension

https://www.typescriptlang.org/docs/handbook/esm-node.html#new-file-extensions


alex-courtis (approved):

PR #304, opened 2023-09-14
feat: add avif filetype

adrian5:
Seeing it in the wild more and more, and using it myself. This format is here to stay.

I've used the properties shared by other image formats.


gegoune:
Thank you!

PR #317, opened 2023-10-05
feat: add Dockerfile extension

steschwa:
Icon for `*.Dockerfile` files

### DIFF
--- a/lua/nvim-web-devicons.lua
+++ b/lua/nvim-web-devicons.lua
@@ -1656,6 +1656,12 @@ local icons_by_file_extension = {
     cterm_color = "106",
     name = "Liquid",
   },
+  ["Dockerfile"] = {
+    icon = "󰡨",
+    color = "#458ee6",
+    cterm_color = "68",
+    name = "Dockerfile",
+  },
 }
 
 -- When adding new icons, remember to add an entry to the `filetypes` table, if applicable.
--- a/lua/nvim-web-devicons-light.lua
+++ b/lua/nvim-web-devicons-light.lua
@@ -1654,6 +1654,12 @@ local icons_by_file_extension = {
     cterm_color = "58",
     name = "Liquid",
   },
+  ["Dockerfile"] = {
+    icon = "󰡨",
+    color = "#2e5f99",
+    cterm_color = "25",
+    name = "Dockerfile",
+  },
 }
 
 return {


lua/nvim-web-devicons.lua
```diff
@@ -1656,6 +1656,12 @@ local icons_by_file_extension = {
     cterm_color = "106",
     name = "Liquid",
   },
+  ["Dockerfile"] = {
```
alex-courtis, line 1659: We don't use specific cases for the filenames or extensions as `get_icon` is not case sensitive.

alex-courtis (changes requested):
What's the use case for this one? Is a `Dockerfile` not correctly applying somewhere?

steschwa:
> What's the use case for this one? Is a `Dockerfile` not correctly applying somewhere?

If you want to build multiple docker images using the same context. Here's an example:

image

I am building an API that is distributed as a docker image. I therefore have a `Dockerfile`. To run database migrations i build another docker image. Because `Dockerfile` is already used i need a different filename `migrations.Dockerfile`. There is no official recommendation (as far as i am concerned) what naming convention to use to name multiple Dockerfiles. I found this SO https://stackoverflow.com/questions/26077543/how-to-name-dockerfiles that suggests naming them with the `.Dockerfile` extension.

lua/nvim-web-devicons.lua
steschwa: This is interesting. Currently i have the following configuration for `nvim-web-devicons` (notice the uppercase "Dockerfile"):

```lua
override_by_extension = {
    ["Dockerfile"] = {
        icon = "󰡨",
        color = "#458ee6",
        cterm_color = "68",
        name = "Dockerfile",
    },
},
```

if i chance it to be lowercased "dockerfile" the below example with `migrations.Dockerfile` doesn't get an icon

alex-courtis:
> Because `Dockerfile` is already used i need a different filename `migrations.Dockerfile`. 

I understand. I did not note that it's an extension rather than a name.

We also need to handle the case of `Dockerfile.foo` which is a pattern used in the wild.

```
dockerfile
Dockerfile
dockerfile.bar
Dockerfile.bar
foo.dockerfile
foo.Dockerfile
foo.dockerfile.bar
foo.Dockerfile.bar
```

alex-courtis (approved):

PR #322, opened 2023-10-08
chore: reorder icon entries alphabetically

kmoschcau:
This would still need the other PR merged in. Also I am not sure whether the icons are case sensitive or not. If they are not, we might also want to remove the upper case ones.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -10,6 +10,10 @@ Please name your commits and the PR simply e.g.
     update makefile icon
     update .kt colors
 
+## Order
+
+Please ensure `icons_by_filename`, `icons_by_file_extension` and `filetypes` are ordered alphabetically, to prevent merge conflicts.
+
 ## Prerequisites
 
 Add [vim-colortemplate](https://github.com/lifepillar/vim-colortemplate) to &runtimepath. The easiest way to do this is via your package manager.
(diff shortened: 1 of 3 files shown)


alex-courtis (approved):
```lua
local log = require "nvim-tree.log"

for n, i in pairs(icons_by_filename) do
  log.raw("dev", "icons_by_filename[%s] %s %s %s %s\n", n, i.icon, i.color, i.cterm_color, i.name)
end

for e, i in pairs(icons_by_file_extension) do
  log.raw("dev", "icons_by_file_extension[%s] %s %s %s %s\n", e, i.icon, i.color, i.cterm_color, i.name)
end
```

```sh
sort ${XDG_STATE_HOME}/nvim/nvim-tree.log > /tmp/icons.master
sort ${XDG_STATE_HOME}/nvim/nvim-tree.log > /tmp/icons.master
: ; diff /tmp/icons.master /tmp/icons.new
274c274
< icons_by_filename[procfile]  #a074c4 140 Procfile
---
> icons_by_filename[procfile]  #a074c4 197 Procfile
```

Procfile was incorrect on master.

Many thanks for this change - it will make additions much easier in the future.

alex-courtis:
> Also I am not sure whether the icons are case sensitive or not. If they are not, we might also want to remove the upper case ones.

I think the extensions like `R` need case sensitivity. There's only a couple...

PR #323, opened 2023-10-10
refactor: move default icon theme to separate file

Akmadan23:
I was browsing the code and I felt like having this kind of separation would be much more intuitive, especially for newcomers. Moreover, having the two themes laid out this way makes comparisons between them with diff tools way easier.


alex-courtis (approved):
Just fantastic!

It is quite the pleasure to work with you @Akmadan23

Akmadan23:
Oh, it's my pleasure to help and be part of this team!

PR #325, opened 2023-10-15
feat: Icons for Python type annotation and Cython source

abelcheung:
More python related entries:

- `py.typed` and `.pyi` are [Python type annotation files](https://peps.python.org/pep-0561/#partial-stub-packages)
- `.pyx`, `.pxi` and `.pxd` are [Cython sources](https://cython.readthedocs.io/en/latest/src/userguide/language_basics.html#cython-file-types). Cython is considered as different language in vim (`filetype=pyrex`), but still belongs to Python family, so bluish color is used instead of yellow in original Python files.

### DIFF
--- a/lua/nvim-web-devicons/icons-default.lua
+++ b/lua/nvim-web-devicons/icons-default.lua
@@ -1243,13 +1243,13 @@ local icons_by_file_extension = {
   ["pxd"] = {
     icon = "",
     color = "#5aa7e4",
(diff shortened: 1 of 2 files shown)


alex-courtis (commented):

Next time please generate the colours as per https://github.com/nvim-tree/nvim-web-devicons/blob/master/CONTRIBUTING.md

PR #326, opened 2023-10-16
feat: add icons for video files

Akmadan23:
The color I chose is `#FD971F` but it's just a proposal, please let me know if you have a different idea for it.


gegoune (approved):
Don't mind the colour. Whatever works.
Thanks!

PR #333, opened 2023-11-11
fix: correct solidity icon

alisinabh:
(This might be a bit opinionated)

Switch to using Ethereum project icon `nf-seti-ethereum` instead of Ethereum currency icon `nf-md-currency_eth` for solidity files.

Ideally, the solidity icon itself should be used but that is not available in the nerd fonts. But the current currency icon used does not make a lot of sense.

Thanks!


alex-courtis (approved):

PR #335, opened 2023-11-14
ci: generate cterm color even if line is missing

Akmadan23:
I noticed that some icons don't have a `cterm_color` value, so I updated the script to generate it if it's missing.

### DIFF
--- a/scripts/generate_colors.lua
+++ b/scripts/generate_colors.lua
@@ -22,6 +22,9 @@ if not rc then
   error(err .. "\nPlease ensure lifepillar/vim-colortemplate is present in the runtimepath.")
 end
 
+-- Needed in order to have the correct indentation on line insertion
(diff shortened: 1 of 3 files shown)


Akmadan23:
This also fixes the bug that caused the assignment of the generated cterm color (if its line was missing) to the next icon

scripts/generate_colors.lua
```diff
@@ -22,6 +22,9 @@ if not rc then
   error(err .. "\nPlease ensure lifepillar/vim-colortemplate is present in the runtimepath.")
 end
 
+-- Needed in order to have the correct indentation on line insertion
+vim.o.autoindent = true
```
alex-courtis, line 26: very nice

alex-courtis (approved):
Works, great catch

PR #338, opened 2023-11-19
ci: harden generation

alex-courtis:
Make started failing unpredictably for me. It could be a result of my setup, as we are running the user's full nvim setup.

* Run with `--clean`
* Clone `lifepillar/vim-colortemplate` if needed
* Fail hard
* Simplify CI to use make

Ran out of energy refactoring the ci further...


Akmadan23 (approved):
Really nice, running the script in a clean environment makes it actually deterministic. Thank you for making this!

PR #341, opened 2023-11-19
ci: split jobs and pin versions

alex-courtis:
Looks more like the nvim-tree CI.

* versions pinned
* uses make targets
* splits into three jobs - yes that starts 3 docker instances however it seems to be best practice
* runs on master

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -29,16 +29,12 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
-      - uses: leafo/gh-actions-lua@v9
-        with:
-          luaVersion: "5.1"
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -0,0 +1,55 @@
...
+
+      - uses: leafo/gh-actions-lua@v10
+        with:
+          luaVersion: "5.1"
+
+      - uses: leafo/gh-actions-luarocks@v4
+
+      - name: luacheck
+        run: |
+          luarocks install luacheck 1.1.1
```
alex-courtis, line 28: This seems to be a lot faster than installing the ubuntu package and allows us to pin the version.

Akmadan23 (approved):
Much cleaner approach, very appreciated!

PR #346, opened 2023-11-26
feat: luau color change and .luaurc

lopi-py:
Nerd fonts still does not have a luau icon, but at least let's keep the original color from the website: https://luau-lang.org/. Luau also supports a `.luaurc` file where you can configure some settings for analysis.


lopi-py:
Uh, what does that fail mean

alex-courtis:
> Uh, what does that fail mean

Please generate the light colour variants as per https://github.com/nvim-tree/nvim-web-devicons/blob/master/CONTRIBUTING.md

`make colors` should do it.

alex-courtis (approved):

PR #349, opened 2023-11-27
ci: remove colors from pre commit

alex-courtis:
@gegoune I think I am getting this wrong.

The 'Pre-commit autoupdate' jobs have been passing, until I moved them to the makefile.

The jobs have been failing as `make colors` is called multiple times against all files. That has been reverted.

However it seems that the colors job has never been passing - it's not had vim-colortemplate in the runtimepath. It is marked as successful. This branch is at the last known good commit, with some rubbish in the job. It's passing: https://github.com/nvim-tree/nvim-web-devicons/actions/runs/7000075636

1. Do we need colors in the pre-commit hook or is CI enough?
2. How can we make them work, via make or other means?

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -11,11 +11,3 @@ repos:
         entry: luacheck --
         language: system
         types: [lua]
-  - repo: local
-    hooks:
-      - id: lighttheme
-        name: Light Color Scheme Generator
-        description: Ensures Light Color Scheme version has been generated.
-        entry: nvim --headless -c 'source scripts/generate_colors.lua' -c 'qall'
-        language: system
-        types: [lua]
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -3,20 +3,19 @@ repos:
     rev: v0.19.1
     hooks:
       - id: stylua-github
-
   - repo: local
     hooks:
       - id: luacheck
         name: Luacheck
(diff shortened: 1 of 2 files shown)


gegoune:
Before we dive into it - do you even use pre-commit? Maybe we should just remove it? I can look into making it work at some point soon if you would like to keep it.

alex-courtis:
> Before we dive into it - do you even use pre-commit? Maybe we should just remove it? I can look into making it work at some point soon if you would like to keep it.

I don't but I have several colleagues who make great use of it.

I'm quite happy to keep it - it's doing all the things that CI does and will make PRs easier.

.pre-commit-config.yaml
```diff
@@ -3,20 +3,27 @@ repos:
     rev: v0.19.1
     hooks:
       - id: stylua-github
-
+        fail_fast: true
```
alex-courtis, line 6: The steps will interfere with each other and cause spurious subsequent failures.

.pre-commit-config.yaml
```diff
@@ -3,20 +3,27 @@ repos:
...
+        types: [lua]
   - repo: local
     hooks:
       - id: colors
         name: colors
         description: Ensures Light Color Scheme version has been generated.
         entry: make colors
         language: system
+        require_serial: true
+        pass_filenames: false
```
alex-courtis, line 27: Single run rather than one per modified file.

PR #350, opened 2023-12-12
feat: update `toml` icons

cpea2506:
Toml icon has been added recently since Nerd Font version `3.1.1`

## Code point:	
Screenshot 2023-12-12 at 23 38 33

## Preview:
Project showcase

## Side note: 
Since the original logo has white background, white color is prefer to the original `#7E7F7F` to make it stand out from others. Toml logo

### DIFF
--- a/lua/nvim-web-devicons/icons-default.lua
+++ b/lua/nvim-web-devicons/icons-default.lua
@@ -1659,9 +1659,9 @@ local icons_by_file_extension = {
     name = "TFVars",
   },
   ["toml"] = {
-    icon = "",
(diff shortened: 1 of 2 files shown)


gegoune:
I am not entirely sure about the colour change. Will merge it now anyway. Thanks!

wookayin:
Ref https://github.com/ryanoasis/nerd-fonts/pull/1436

samgj18:
I think this change cause a regression. Is not working for all fonts. This is how it looks for Fira Code Nerd (which should support all the icons). Previous icon was working properly

image

Issue #361, opened 2023-12-25
Are you open to a PR for adding operating system icons?


miversen33:
I know that devicons is traditionally used for showing the icons of files, but I am looking for an easy way to show the icons of an operating system.

Per [this comment on #155](https://github.com/nvim-tree/nvim-web-devicons/issues/300#issuecomment-1704038616), nvim-web-devicons can use icons provided by nerdfonts.

Nerdfonts provides several icons for popular operating systems and linux distros

It would be pretty slick if we could include these icons as well.

If the team is open to it, I would happily contribute this

alex-courtis:
Yes please [Contributions are always welcome](https://github.com/nvim-tree/nvim-web-devicons/blob/master/CONTRIBUTING.md)

I imagine we would have to use the distro name as the extension and name. Users could request the icon by distro.

hasecilu:
Maybe this should be closed as #433 included all the icons of OSs from Nerd Fonts - Font Logos.

alex-courtis:
> Maybe this should be closed as #433 included all the icons of OSs from Nerd Fonts - Font Logos.

Thank you.

You should have permissions to close issues etc.

Please let me know if you can't close this one!

hasecilu:
Yes I can, I just wanted to discuss first, but yeah, for this issue the result is pretty straight forward.

alex-courtis:
> Yes I can, I just wanted to discuss first, but yeah, for this issue the result is pretty straight forward.

Excellent. You have my trust.

PR #362, opened 2023-12-26
feat: operating system icons

miversen33:
#361 

There are a few things to call out here.

- This adds a completely new table to the return of `icons-default` and `icons-light` called `icons_by_operating_system`. This set of icons is added with the other icons to the lookup table that is generated during initialization of all devicons.
- This (currently) does not pass the provided make tests due to the addition of the new `icons_by_operating_system` attribute
- There are several icons that I am choosing to omit from this. The following list of icons were omitted as the provided icon by [nerdfonts cheat-sheet](https://www.nerdfonts.com/cheat-sheet) was not found when I pasted it into my terminal. I am currently using the patched version of [firacode](https://github.com/tonsky/FiraCode) so I suspect these various items are all in a newer patch for Firacode. In any case, I did not see them and thus did not want to commit them. I did complete them however so a lua file is attached to this message with each one so that it can be reviewed should someone decide to include these icons
  - archcraft
  - arcolinux
  - arduino
  - biglinux
  - crystal
  - garuda
  - hyperbola
  - kde_neon
  - kubuntu
  - locos
  - lxle
  - mate
  - mxlinux
  - parabola
  - puppy
  - qubesos
  - tails
  - trisquel
  - vanilla
  - xerolinux
  
Pretty much all colors were established via "best guess"/"eyeball test". Others have figured out Apple/Windows color palettes and thus for these 2 I pulled the primary color from their established color palette. Feel free to adjust the color codes as needed. Color can be subjective after all :)

On colors, I did not do anything fancy for any of the OS icons for the 
"light" vs "default". It is a literal copy and paste between the 2. If someone wants different colors for the OS when requesting a "light" version of them, they can do that work lol.

I know that this PR doesn't exactly follow the Contributing guidelines request but given that is a bit chunky and adding stuff that doesn't _really_ have a home, I figured it would be worth explaining that. Do let me know what you believe needs to be changed/fixed/etc to get this in :)

[potential-icons.zip](https://github.com/nvim-tree/nvim-web-devicons/files/13767590/potential-icons.zip)

### DIFF
--- a/lua/nvim-web-devicons.lua
+++ b/lua/nvim-web-devicons.lua
@@ -1,7 +1,7 @@
 local M = {}
 
 -- When adding new icons, remember to add an entry to the `filetypes` table, if applicable.
-local icons, icons_by_filename, icons_by_file_extension
+local icons, icons_by_filename, icons_by_file_extension, icons_by_operating_system
 
 function M.get_icons()
   return icons
@@ -18,7 +18,8 @@ local function refresh_icons()
 
   icons_by_filename = theme.icons_by_filename
   icons_by_file_extension = theme.icons_by_file_extension
-  icons = vim.tbl_extend("keep", {}, icons_by_filename, icons_by_file_extension)
+  icons_by_operating_system = theme.icons_by_operating_system
+  icons = vim.tbl_extend("keep", {}, icons_by_filename, icons_by_file_extension, icons_by_operating_system)
 end
 
 -- Map of filetypes -> icon names
--- a/lua/nvim-web-devicons/icons-default.lua
+++ b/lua/nvim-web-devicons/icons-default.lua
@@ -1907,6 +1907,394 @@ local icons_by_operating_system = {
     cterm_color = "33",
     name = "Windows"
   },
+  ["linux"] = {
+    icon = "",
+    color = "#000000",
+    cterm_color = "0",
+    name = "Linux"
+  },
+   ["almalinux"] = {
+     icon = "",
+     color = "#000000",
+     cterm_color = "15",
+     name = "Almalinux"
+   },
+   ["alpine"] = {
+     icon = "",
+     color = "#0d597f",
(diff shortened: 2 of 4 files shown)


alex-courtis:
This looks great. The operating system distinction is not exposed to the user.

I'm not sure about removing linux from the names. The font names all have linux in them and it will be useful to disabmiguate.

@gegoune ? @Akmadan23 ?

miversen33:
> This looks great. The operating system distinction is not exposed to the user.
> 
> I'm not sure about removing linux from the names. The font names all have linux in them and it will be useful to disabmiguate.
> 
> @gegoune ? @Akmadan23 ?

The initial commit had certain distros that had names like `almalinux` or `rocky linux` while others (such as `redhat`) did not. The removal of "linux" from the names was a way to make the names more consistent across the board. IMO they should either all have `linux` in their name, or none of them should in order to prevent confusion and make it easier on people using the icons in scripts/plugins

Akmadan23:
> This looks great. The operating system distinction is not exposed to the user.
> 
> I'm not sure about removing linux from the names. The font names all have linux in them and it will be useful to disabmiguate.
> 
> @gegoune ? @Akmadan23 ?

Looks good. As @miversen33 said, either all or none should have linux in the name, and I'd go with none, since all of them are linux distros (except of course windows, macos and the BSDs, which are pretty obvious).

We should also update the color generation script to handle the `icons_by_operating_system` section now...

miversen33:
I attempted to add the `icons_by_operating_system` and while it "looked good", it seems to break the `icons_by_file_type` table in `icons-light` during the generation. Thus I am going to have to let someone else do that

Akmadan23:
It should be ok now. Could you please run it via `make colors` and push the updates?

miversen33:
Done :)

Akmadan23:
Awesome!

alex-courtis (approved):

PR #364, opened 2023-12-27
fix: scm color

chrisgrieser:
the icon for `.scm` files is black in dark mode. Due to the low contrast, this makes it hard to read. 

This PR changes the color to a bright grey in dark mode (and dark grey in light mode, due to the color generating script) to fix that issue.


alex-courtis (approved):

PR #365, opened 2023-12-28
feat: tsconfig icon

ribru17:
Adds the Nerdfont `nf-seti-tsconfig` icon to `tsconfig.json` files.


alex-courtis (approved):

PR #368, opened 2023-12-30
fix: update cjs & mjs icons

ronitkrshah:
Update **cjs**(CommonJS) & **mjs**(ModuleJS) icons same as **js**(JavaScript) from official Nerd Font icon : `f031e`


alex-courtis (approved):

Issue #287, opened 2023-07-21
fix: persist overrides* and apply during &background change refresh


mikesmithgh:
When you `set background=light` the icon overrides are lost.

See https://github.com/nvim-tree/nvim-web-devicons/blob/master/lua/nvim-web-devicons.lua#L1607

Demo:

https://github.com/nvim-tree/nvim-web-devicons/assets/10135646/73132ff3-7b74-4b46-9fbc-5a9ea54137eb

This doesn't seem likely to happen in a normal session so probably a low priority issue.

gegoune:
There is currently no way to set overrides for dark and light backgrounds independently. Seems like it's a missing crucial functionality.

Bekaboo:
Any updates on this? It's pretty annoying when users switches between dark and light mode and it happens to be everyday.

alex-courtis:
That is not a use case I have ever considered: how do you use background switching as part of your workflow?

We can add an option to disable listening for `&background` changes.

Bekaboo:
@alex-courtis Hi thanks for your reply.

> how do you use background switching as part of your workflow?

I use tmux to for project management, ususally I keep several neovim instances in different tmux sessions for days or even weeks without closing them as long as I am still working on that project.

My desktop switches between light and dark mode automatically according to the time and my neovim is configured to follow these changes, so every time the background changes all my custom icons and colors are lost. To get those back the only way is to restart neovim, which is cumbersome when you have over five neovim instances in tmux sessions.

> We can add an option to disable listening for &background changes.

If we disable changing icon colors for &background changes, I am afraid the icon color will look off in the new background, as mentioned here https://github.com/nvim-tree/nvim-web-devicons/issues/193

ribru17:
> If we disable changing icon colors for &background changes, I am afraid the icon color will look off in the new background, as mentioned here #193

I agree with @Bekaboo, this seems like not the right approach. Perhaps we could store overrides in some variable that persists past the `setup` function? Looks like `global_opts` has an `overrides` field that is never used, I think it would make sense to store user overrides here so they can be properly refreshed, perhaps somewhere here:
```lua
function M.refresh()
  refresh_icons()
  M.set_up_highlights(true)
  -- refresh user overrides stored in global_opts.overrides
end
```

alex-courtis:
I understand, thank you for looking into this @Bekaboo @ribru17 

[...]

PR #369, opened 2023-12-31
fix(#261 #287): persist icon overrides when changing &background

### LINKED ISSUE #261: Overridden icon gets reset on colorscheme change
Overridden icons get reset when I change colorscheme.

**To reproduce:**

Configuration:

<details>
<summary><b>Click to expand</b></summary>

```lua
local root = vim.fn.fnamemodify(vim.fn.expand('$HOME') .. 'nvim-test-config', ':p')

for _, name in ipairs({ 'config', 'data', 'state', 'cache' }) do
  vim.env[('XDG_%s_HOME'):format(name:upper())] = root .. '/' .. name
end

-- bootstrap lazy
local lazypath = root .. '/plugins/lazy.nvim'
if not vim.loop.fs_stat(lazypath) then
  vim.fn.system({ 'git', 'clone', '--filter=blob:none', 'https://github.com/folke/lazy.nvim.git', lazypath, })
end
vim.opt.runtimepath:prepend(lazypath)

-- install plugins
local plugins = {
  { 'romgrk/barbar.nvim', config = true },
  {
    'kyazdani42/nvim-web-devicons',
    config = function()
      require('nvim-web-devicons').setup({
        override = {
          md = { icon = '', color = '#519aba', name = 'Markdown' }
        }
      })
    end,
  }
}

require('lazy').setup(plugins, {
  root = root .. '/plugins',
})
```

</details>

1. `nvim -u ~/nvim-test-config/init.lua testfile.md`
2. Notice the custom markdown icon in the top left
3. `:colorscheme blue`
4. `:e`
5. Markdown icon is back to the default

ribru17:
Fixes #261, #287

> [!NOTE]
> This PR does not address the problem of overrides being unaware of light/dark modes. They are applied universally still. This functionality will have to be implemented later, but this PR is better than nothing.

### DIFF
--- a/lua/nvim-web-devicons.lua
+++ b/lua/nvim-web-devicons.lua
@@ -7,6 +7,13 @@ function M.get_icons()
   return icons
 end
 
+local global_opts = {
+  override = {},
+  strict = false,
+  default = false,
+  color_icons = true,
+}
+
 -- Set the current icons tables, depending on the 'background' option.
 local function refresh_icons()
   local theme
@@ -20,6 +27,7 @@ local function refresh_icons()
   icons_by_file_extension = theme.icons_by_file_extension
   icons_by_operating_system = theme.icons_by_operating_system
   icons = vim.tbl_extend("keep", {}, icons_by_filename, icons_by_file_extension, icons_by_operating_system)
+  icons = vim.tbl_extend("force", icons, global_opts.override)
 end
 
 -- Map of filetypes -> icon names
@@ -229,13 +237,6 @@ local default_icon = {
   name = "Default",
 }
 
-local global_opts = {
-  override = {},
-  strict = false,
-  default = false,
-  color_icons = true,
-}
-
 local function get_highlight_name(data)
   if not global_opts.color_icons then
     data = default_icon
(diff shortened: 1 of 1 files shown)


Bekaboo:
> This PR does not address the problem of overrides being unaware of light/dark modes. They are applied universally still. This functionality will have to be implemented later, but this PR is better than nothing.

It will be great if we can just keep the current config structure and auto-generate the dark version (used with light bg) of the icons provided by users using the same logic that generates the dark icons for the default ones. I would like to help with this when I have time.

ribru17:
I think ideally this should be done in another PR, I think getting this one merged on it's own will be good enough for many cases. From there I could try to make a PR for what you are saying. I could be wrong but it looks like the light mode icons are generated from the dark mode ones, given the darkening function in the scripts. Maybe we can just expose it publicly (it is private right now) and then use it directly in the refreshing function, detecting when the background is `light`?

Bekaboo:
Yeah I mean I can make another PR for the problem of overrides being unaware of light/dark modes. Haven't gotten time to really look into this though, it will be cool if you are willing to do it!

ribru17:
I think this PR is good to merge. I have a change waiting that will also update the highlights properly, but only for hex colors, not cterm. I'm not sure how to handle that one.

alex-courtis (approved):
Works nicely, many thanks for your contribution!

PR #370, opened 2023-12-31
feat(#168): add prefixes

### LINKED ISSUE #168: Add icon for Dockerfile or dockerfile with extensions

BBboy01:
Closes #168

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,9 +65,9 @@ require'nvim-web-devicons'.setup {
  -- will get overriden by `get_icons` option
  default = true;
  -- globally enable "strict" selection of icons - icon will be looked up in
- -- different tables, first by filename, and if not found by extension; this
+ -- different tables, first by filename, then by prefix, and if not found by extension; this
  -- prevents cases when file doesn't have any extension but still gets some icon
- -- because its name happened to match some extension (default to false)
+ -- because its name happened to match some prefix or extension (default to false)
  strict = true;
  -- same as `override` but specifically for overrides by filename
  -- takes effect when `strict` is true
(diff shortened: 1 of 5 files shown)


BBboy01:
`make colors` makes me delete `icons_by_file_prefix` table in `lua/nvim-web-devicons/icons-light.lua`, but this will occur an exception no icons show after set `vim.o.background = "light"`, how can i fix this? If there are something i missed, let me know plz.

README.md
```diff
@@ -92,12 +102,12 @@ require'nvim-web-devicons'.setup {
...
 ### Get Icon
 
-Get the icon for a given file by passing in the `name`, the `extension` and an _optional_ options `table`.
+Get the icon for a given file by passing in the `name`, the `pre_or_ext` and an _optional_ options `table`.
 The name is passed in to check for an exact match e.g. `.bashrc` if there is no exact name match the extension
 is used. Calls `.setup()` if it hasn't already ran.
 
 ```lua
-require'nvim-web-devicons'.get_icon(filename, extension, options)
+require'nvim-web-devicons'.get_icon(filename, pre_or_ext, options)
```
alex-courtis, line 110: This will break existing behaviour - a prefixed icon will be returned where none was before.

lua/nvim-web-devicons.lua
```diff
@@ -428,9 +481,15 @@ function M.get_icon_colors(name, ext, opts)
...
   if is_strict then
-    icon_data = icons_by_filename[name] or get_icon_by_extension(name, ext, opts) or (has_default and default_icon)
+    icon_data = icons_by_filename[name]
+      or get_icon_by_prefix(name, pre_or_ext, opts)
+      or get_icon_by_extension(name, pre_or_ext, opts)
+      or (has_default and default_icon)
   else
-    icon_data = icons[name] or get_icon_by_extension(name, ext, opts) or (has_default and default_icon)
+    icon_data = icons[name]
+      or get_icon_by_prefix(name, pre_or_ext, opts)
```
alex-courtis, line 490: This will result in a prefix being returned when users were previously expecting an extension, when the key is the same.

alex-courtis (changes requested):
I've gone cold on this one; it's introducing breaking changes.

We need to refactor this change - specifically the API - such that existing behaviour does not change. For the user to return the prefix icon they will need to explicitly specify it.

Unfortunately we haven't used named parameters for all API methods however we cannot change that. `options` is available in some cases but not all.

Proposal:
Draft an API that preserves existing behaviour and allows prefixes to be selected.

Extension:
New API that uses all name parameters via `options`

alex-courtis:
Please reopen if you wish to continue this change.

Issue #192, opened 2023-01-26
Proposal: Use Nerd Font css class names instead of hardcoding


mochaaP:
# Background

Nerd Fonts plans to introduce [a breaking change](https://github.com/ryanoasis/nerd-fonts/discussions/1074) in v3.0.0:

> ## Next Release Outlook
> 
> The `v3.0.0` release will bring two bigger changes:
> 
> ### Material Design Icons
> The Material Design Icons have to move codepoints. Historically we patch them in between `F500` - `FD46`; but part of that range is needed for Chinese and other glyphs. With #773 and _Plan 5+_ of #365 their new home will be after `F0001`. And they grow considerably.
> 
> The `v2.3.x` release has both, the 'old' and the 'new' codepoints to enable you an easier transision. **Please change** already **now** your prompts and other uses to the new codepoints. Release `v3.0.0` will finally drop the old codepoints (`F500` - `FD46`).
> 
> ### Naming
> Naming will also change. The name parts will move to places where they are customarily expected (e.g. weight and style comes last). This will fix some more problems. Probably we will drop the `Complete` monicker, as ... everyone uses the complete versions, it just takes precious space in the names.
> The details are not fixed yet, follow [#1074](https://github.com/ryanoasis/nerd-fonts/discussions/1074) to keep up to date.

Hardcoding glyphs are very fragile to those changes. Moreover, if a user choose to use custom codepoint ranges (due to conflicts, etc.), it's a hassle to manually replace all icons in every plugin.

# Solution

Fortunately, Nerd Fonts do have methods for getting glyphs via name, one via CSS class names, one being [work in progress](https://github.com/ryanoasis/nerd-fonts/issues/448#issuecomment-1405132215).

I suggest we could made the glyph table into a single .lua file and reference that and additionally, sync upstream with CI.

alex-courtis:
That sounds fantastic, more robust references are always desirable.

Perhaps you could create a sample lua file with the table. We can plan this change from there.

mochaaP:
I made a quick script that always returns the latest glyph data as JSON at <https://nerd-fonts-glyph.deno.dev/>. Hope that helps :-)

alex-courtis:
That's great. I'll see what I can do with it.

Not sure how we'll handle 2/3 compatibility; any guidance?

mochaaP:
Only MDI's glyph will be moved. Other icons are fine.

alex-courtis:
Thank you!

[...]

PR #376, opened 2024-01-12
fix: update ts, cts, mts and tsconfig.json icon

ofseed:
The original ts and js icons keeps the same style:

image

But https://github.com/nvim-tree/nvim-web-devicons/commit/808627b8d412b2a6b6fc6eed816fec3557198b01 and https://github.com/nvim-tree/nvim-web-devicons/commit/33e27b859f549873a826d8aaa348fd31195e439f break this:

image

So I recommend to update the ts icon as well:

image

And use the original ts icon for `tsconfig.json`, because they are very similar.

image

**But what I recommend most is to revert these two PRs directly.** There is no reason to do this update, and this also affects the uniform style of other icons, for test.js

image

This new icon is a completely different style:

image

Note: 
I'm not swapping icons for tsconfig and ts, These two icons look the same but they have very subtle differences. These are two icons with different code points. Compare to ts:

e69d(nf-seti-tsconfig):
image

f06e6(nf-md-language_typescript):
image

### DIFF
--- a/lua/nvim-web-devicons/icons-default.lua
+++ b/lua/nvim-web-devicons/icons-default.lua
@@ -336,7 +336,7 @@ local icons_by_filename = {
     name = "SvelteConfig",
   },
   ["tsconfig.json"] = {
-    icon = "",
+    icon = "",
     color = "#519aba",
     cterm_color = "74",
     name = "TSConfig",
@@ -615,7 +615,7 @@ local icons_by_file_extension = {
     name = "Csv",
   },
   ["cts"] = {
-    icon = "",
+    icon = "󰛦",
     color = "#519aba",
     cterm_color = "74",
     name = "Cts",
@@ -1203,7 +1203,7 @@ local icons_by_file_extension = {
     name = "Mp4",
   },
   ["mts"] = {
-    icon = "",
+    icon = "󰛦",
     color = "#519aba",
     cterm_color = "74",
     name = "Mts",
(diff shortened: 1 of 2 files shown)


alex-courtis:
Thank you for finding all these inconsistencies in the various script icons.

The test icons were added #215 based on vscode and are not "official" javascript icons.

alex-courtis:
We appear to be in a loop here, moving between icon sets.

Seti and material icons are present however we are inconsistent.

VSCode default seti icons:


VSCode material icons:

alex-courtis:
Idea: audit the icons noting the class and hex.

This could be done during generate_colors.lua and comments added to the icon.

gegoune:
Maybe it's time to introduce icon sets? Or pick one style and support only that one. Consistency is a good thing.

alex-courtis:
> Maybe it's time to introduce icon sets? Or pick one style and support only that one. Consistency is a good thing.

Yes. We seem to mainly use seti, octicons, devicons and material.

Let's see how an audit goes.

alex-courtis:
Some analysis, mapping the icons to their nerd-font classes:

https://gist.github.com/alex-courtis/26271d581e390fe26058bf0743539cc3

TL;DR: we are using mostly seti and dev with some material design and font awesome.

> Maybe it's time to introduce icon sets? Or pick one style and support only that one. Consistency is a good thing.

I'm happy to pick seti and use that for the defaults.

Users can add different icon sets via setup and I think there is a material configuration out there somewhere. A contributor could add icon sets with plumbing to nvim-web-devicons if they are motivated.

It might be time to reopen #192 which will be simpler now that we have the generation scripts.

alex-courtis:
Many thanks @ofseed and all for the great discussion.

I've updated #192 to move to css class names.
#391 will then consolidate on seti icons
#392 will add a material set

I'll close this and #377 - we will resolve all the js/ts icons in one change.

PR #379, opened 2024-01-13
fix: `get_icons()[1]` being nil after changing &background

Bekaboo:
The default icon should be in `get_icons()[1]` but it disappears after changing the background. This PR fixes the issue.


alex-courtis (approved):

PR #381, opened 2024-01-15
feat: Add README icons

DreepyYunky:
READMEs (eg `README.md`) are special files, so giving them special icon can help distinct them from, for example, an average markdown file.

The icon is `nf-md-information` which I think fits since READMEs give information about the thing it's on.


alex-courtis:
Holding off pending #376

DreepyYunky:
> Holding off pending #376

Ok.

pauchiner:
@DreepyYunky i think you should also add for `.rst` README files, these are pretty common in python enviroments

DreepyYunky:
> @DreepyYunky i think you should also add for `.rst` README files, these are pretty common in python enviroments

I even might add for files just named `README`, no txt or md.

DreepyYunky:
@pauchiner Added `.rst`s now.

alex-courtis:
Sorry, I don't think we should do this: markdown and text are file types, readme is not. web-devicons only handles types, particularly those specified by languages.

That's kind of confirmed by the lack of readme icon in nerd fonts ;)

If you desire special file name handling you can use API `override_by_filename`

PR #393, opened 2024-01-29
feat: add C++20 extentions

### LINKED ISSUE #390: Icons for C++20 module files doent work
C++20 introduced modules and different compilers have different file extensions:
MSVC: .ixx
Clang: .cppm .cxxm .ccm
GCC: .mpp

It would be nice to at least recognize files with these extensions as C++ files and add appropriate icons

neg-c:
fixes #390


alex-courtis (approved):
Matches other c++ icons and colours.

We can look at seti at a later date.

PR #394, opened 2024-01-29
feat: add haxe

bwkam:
tried to add a [haxe](https://haxe.org/) icon, still isn't showing for me though


alex-courtis (approved):
Seti icons.

Many thanks for your contrtibution!

PR #401, opened 2024-02-16
feat: add nu

steschwa:
Icon for Nushell files https://www.nushell.sh/

- Color is Nushell's "Brand" color
- Icon is generic because Nerd Fonts don't include Nushell's chevron-like logo

### DIFF
--- a/lua/nvim-web-devicons.lua
+++ b/lua/nvim-web-devicons.lua
@@ -157,6 +157,7 @@ local filetypes = {
   ["mustache"] = "mustache",
   ["nim"] = "nim",
   ["nix"] = "nix",
+  ["nu"] = "nu",
   ["node"] = "node_modules",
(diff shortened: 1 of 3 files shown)


alex-courtis:
Octo has nf-oct-chevron_right `` 

Could we use that instead of nf-oct-command_palette ``?

Perhaps we could even just use regular greater than `>`

steschwa:
> Octo has nf-oct-chevron_right ``
> 
> Could we use that instead of nf-oct-command_palette ``?
> 
> Perhaps we could even just use regular greater than `>`

Sure! i tried `nf-oct-chevron_right` but i think a simple `>` looks cleaner because it does not overflow a single cell (unlike some of the `nf-oct-*` icons)

alex-courtis (approved):
Beautiful! Thank you for your contribution.

PR #410, opened 2024-03-03
feat: add audio file icons, remove audio/video from vim filetypes

### LINKED ISSUE #404: Add mp3 file icons
### Current Behaviour

Captura de pantalla 2024-02-22 a las 14 24 09

I think that any of these icons would be a great fit in this case:

Captura de pantalla 2024-02-22 a las 14 24 46

pauchiner:
Closes #404

Preview

Roadmap:
- [X] .aac (Advanced Audio Coding)
- [X] .aiff or .aif (Audio Interchange File Format)
- [x] .ape (Monkey's Audio)
- [X] .flac (Free Lossless Audio Codec)
- [X] .mp3 (MPEG-1 Audio Layer)
- [X] .m4a (MPEG-4 Audio)
- [X] .opus (Opus Audio File)
- [X] .ogg (Ogg Vorbis)
- [X] .wav (Waveform Audio File Format)
- [X] .wma (Windows Media Audio)


alex-courtis (changes requested):
Looking good; let's add the rest https://github.com/nvim-tree/nvim-web-devicons/issues/404#issuecomment-1975375731

alex-courtis (changes requested):
These are looking good, however they should actually be in the `icons_by_file_extension` table.

There are already some of these extensions in there... including mp3.

I'll be grateful if you could move the new ones in there, overwriting.

However... the existing ones are using `nf-fa-music`. We'll need to use that one, to avoid a breaking change.

pauchiner:
Sorry, I made those commits in a few free minutes yesterday, I didn't notice either the `extensions table` or the `breaking changes`, I'm already working on fixing it.

alex-courtis:
Looking good: `:NvimWebDeviconsHiTest`

However I think we might be backwards here: all the new icons are in `icons_by_filename` wheras they should be in `icons_by_file_extension`. The user will be finding music file icons via `get_icon` with an extension rather than matching an exact file name.

The `filetypes` will still correctly map to `icons_by_file_extension`.

alex-courtis:
Any updates @pauchiner ? This is really close...

pauchiner:
Sorry, I was absent due to personal things, as soon as I can I will make a commit with the changes.

lua/nvim-web-devicons/icons-default.lua
```diff
@@ -1,4 +1,10 @@
 local icons_by_filename = {
+  ["wma"] = {
+    icon = "",
+    color = "#00afff",
+    cterm_color = "39",
+    name = "WindowsMediaAudio",
+  },
```
alex-courtis, line 7: Please remove - it's present in extensions.

lua/nvim-web-devicons.lua
```diff
@@ -246,6 +249,7 @@ local filetypes = {
   ["vue"] = "vue",
   ["wasm"] = "wasm",
   ["wav"] = "wav",
+  ["wma"] = "wma",
```
alex-courtis, line 252: The above aren't vim filetypes so should not be mapped to extensions.

alex-courtis (changes requested):
Looking good, a couple of tidies needed.

alex-courtis (changes requested):
Still not quite there...

None of these file types correspond to vim syntax filetypes, see `/usr/share/nvim/runtime/syntax`

We need to remove them from the mapping
```lua
-- Map of filetypes -> icon names
local filetypes = {
```

pauchiner:
So if aren't vim filetypes, how are mapped?
I think im not understanding your point, Sorry.

alex-courtis:
> So if aren't vim filetypes, how are mapped? I think im not understanding your point, Sorry.

No worries; this is not very clear from the names.

Files like these are simply mapped by their extension against the keys of `icons_by_file_extension`. There's no need for any extra mapping.

Try
```
:lua print(require("nvim-web-devicons").get_icon("aiff"))
```

hasecilu:
I'm not really sure if [PCM](https://en.wikipedia.org/wiki/Pulse-code_modulation) file extension is really used, so could be added.

pauchiner:
I have already deleted the filetypes that I added at the time, but then by the same logic, those that were previously there should also be deleted, such as mp3, mp4, wav...

alex-courtis:
> I have already deleted the filetypes that I added at the time, but then by the same logic, those that were previously there should also be deleted, such as mp3, mp4, wav...

Good catch! Yes, they should be removed.

In the interest of speed, I'll push that change to this branch.

alex-courtis:
> I'm not really sure if [PCM](https://en.wikipedia.org/wiki/Pulse-code_modulation) file extension is really used, so could be added.

Sounds good! Any others you think of please do raise a PR...

alex-courtis (approved):

alex-courtis (approved):
Many thanks for your contribution and many thanks for bearing with all the back and forth.

Issue #391, opened 2024-01-29
Consolidate Default Icon Set: Seti


alex-courtis:
We currently have a mix of [icon sets](https://gist.github.com/alex-courtis/26271d581e390fe26058bf0743539cc3?permalink_comment_id=4829827#gistcomment-4829827)

```
      2 fae-
      7 NOGLYPH-
      8 cod-
     13 oct-
     27 custom-
     31 linux-
     32 fa-
     58 md-
     76 dev-
    122 seti-
```

Seti is the majority and should be used wherever possible.

#192 will ease this change.

alex-courtis:
There are only 167 seti icons. We could add them all.

```sh
: ; cat glyphnames.json | jq -r 'keys[]' | sed -E 's/([^-])-.*/\1/g' | sort | uniq -c
    387 cod
     33 custom
    191 dev
    786 fa
    170 fae
      5 iec
      2 indent
      1 indentation
    115 linux
   6880 md
      1 METADATA
    310 oct
      9 pl
     32 ple
     11 pom
    167 seti
    228 weather
```

PR #414, opened 2024-03-09
refactor: rewrite generator

futsuuu:
Currently, icons are defined separately one by one, so it takes time to add or change icons for the same type of file.
For example, if we want to add more icons of same format, we have to repeat the copy and paste for each extension. After a while, if we want to change the colors of these icons, we will have to find all of them and replace the colors correctly.

This PR rewrites the generator to fix these probrem.

before:

```diff
+ [".gitattributes"] = { icon = ".", color = "#......", name = "GitAttributes" },
  [".gitconfig"] = { icon = ".", color = "#......", name = "GitConfig" },
+ [".gitkeep"] = { icon = ".", color = "#......", name = "GitKeep" },
+ [".gitmodules"] = { icon = ".", color = "#......", name = "GitModules" },
```

after:

```diff
- Git = { icon = "seti-git", color = "#......", file = { ".gitconfig" } },
+ Git = { icon = "seti-git", color = "#......", file = { ".gitattributes", ".gitconfig", ".gitkeep", ".gitmodules" } },
```


alex-courtis:
This looks great! It works and will be a dev time savings.

This is a refactor of the existing functionality: a major refactor of the source files. That adds value.

This is, however, a nearly full rewrite of the generator. Unfortunately I can't justify such a large change, with the additional complexity and maintenance it requires.

alex-courtis:
All is not lost!

We have a major generation change planned: use the css classes rather than icons: #192 which would allow us to create icon sets, starting with seti #391

This css class change would marry beautifully with this refactor. Are you up for the challenge @futsuuu ?

alex-courtis:
We will help you make and test this change. See https://github.com/nvim-tree/nvim-web-devicons/pull/376#issuecomment-1890843503 for a proof of concept gist to extract the icons from nerd fonts.

futsuuu:
Sure! But I can't imagine how you want to create icon sets, could you tell me?

futsuuu:
> This is, however, a nearly full rewrite of the generator. Unfortunately I can't justify such a large change, with the additional complexity and maintenance it requires.

I understand this, but the definition and the generated should be separate files for maintenance.

alex-courtis:
> I understand this, but the definition and the generated should be separate files for maintenance.

Absolutely. This change and the addition of css class names does justify the change.

I was being too dramatic in my language...

alex-courtis:
> Sure! But I can't imagine how you want to create icon sets, could you tell me?

Let's build one default sample - seti - first, then we can decide about icon sets.

It may be best to add icon sets after this change.

alex-courtis:
Here are the changes to generate the icons from css classes: #418 

It is clunky and tightly coupled right now. Integration with this rewrite will result in an elegant solution.

hasecilu:
Me at #433 
> Would be possible to create groups, for example create a single icon entry and use it various times for different file extensions to simplify the code? 

This would be very useful because I reuse a lot of icons

alex-courtis:
Handled by #550

gegoune:
> Handled by #550 

No, it's not, unfortunately. This PR aimed to solve the problem of having to define the same icon table repeatedly  when assigning to multiple targets; Instead of defining list of targets with single icon definition. 

Some sort of generator like this would be very valuable, be it what was proposed in this PR, or better yet, autogenerated from css classes (or similar). 

Might be worth reopening although looks like @futsuuu has lost their interest.

alex-courtis:
Ah I see, that would be useful for the repeated ones.

Maybe with CSS ?

PR #428, opened 2024-03-24
feat: add sc

suliatis:
Add icon for scala script files, using the existing scala icon.


alex-courtis (approved):

PR #437, opened 2024-04-07
fix(#435): tex icon

### LINKED ISSUE #435: TeX icon incorrect
Currently, the TeX file type maps to this icon:

image

But I have checked the cheatsheet and the icon mapping seems not right.

image

The correct icon should be this:

image

I don't know if this also happens on others' computers or just mine.

chardoncs:
Update tex icon from 󰙩 to 
fixes #435


chardoncs:
> Many thanks for your contribution!

No problem :)

PR #445, opened 2024-04-18
fix: get_icon_color() is now case insensitive

mcauley-penney:
Problem: `get_icon_color()` does not find icons for files whose names are capitalized and which don't have extensions, such as "LICENSE", "License", or "COMMIT_EDITMSG". This is inconsistent with the functionality of `get_icon()`, which does lowercase the name parameter.

Solution: Lowercase the name parameter in get_icon_color(). Coincidentally, get_icon() and get_icon_color() have a lot of duplicate functionality at the front of the function, specifically to get the entire icon data. Abstract this functionality out to a local function that can be shared by these other two functions, which are now wrapper functions that return different data.

### DIFF
--- a/lua/nvim-web-devicons.lua
+++ b/lua/nvim-web-devicons.lua
@@ -461,7 +461,7 @@ local function get_icon_by_extension(name, ext, opts)
   return iterate_multi_dotted_extension(name, icon_table)
 end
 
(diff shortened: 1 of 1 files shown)


alex-courtis (approved):
Nice straightforward refactor, many thanks for your contribution!

PR #459, opened 2024-04-30
fix(#456): adding gulpfile/gruntfile extensions for js and other js types

### LINKED ISSUE #456: Gulpfile and Gruntfile incorrect icon
image

I'm using Lazy as my package manager devicons is a dependency for these following plugins that I have downloaded lualine, telescope and nvim-tree.
I already did a set_icon() for gulp I was able to change the icon but I wasn't able to change the color. 
image
this is the result when I do a get_icon and a print
image
I already tried to make devicons load faster by putting it in a after/plugins folder which did make it load ahead of the plugins that are dependant on devicons but that didn't fix my problem, the icon is still the JS icon.

RafaelAntonny:
resolves #456, now gulpfile/gruntfile.js, .ts, .babel.js, .coffee will have their proper icons instead of standard JS icon

demo:
nvim-tree
image
telescope
image
bufferline
image
should work across all plugins that have nvim-web-devicons as dependencies

### DIFF
--- a/lua/nvim-web-devicons/icons-default.lua
+++ b/lua/nvim-web-devicons/icons-default.lua
@@ -380,13 +380,13 @@ local icons_by_filename = {
   ["gruntfile.babel.js"] = {
     icon = "",
     color = "#e37933",
-    cterm_color = "167",
+    cterm_color = "166",
     name = "Gruntfile",
   },
   ["gruntfile.coffee"] = {
     icon = "",
     color = "#e37933",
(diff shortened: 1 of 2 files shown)


hasecilu:
Looks good, only thing missing is to execute `./make` on the repository root to generate the light colors.

RafaelAntonny:
done
image

alex-courtis (approved):

PR #468, opened 2024-06-01
feat: add Justfiles

chrisgrieser:
`just` accepts `justfile` and `.justfile`, therefore adding both. https://just.systems/man/en/chapter_57.html?highlight=.justfile#hiding-justfiles

### DIFF
--- a/lua/nvim-web-devicons/icons-default.lua
+++ b/lua/nvim-web-devicons/icons-default.lua
@@ -107,6 +107,12 @@ local icons_by_filename = {
     cterm_color = "28",
     name = "Gvimrc",
   },
+  [".justfile"] = {
+    icon = "",
+    color = "#6d8086",
+    cterm_color = "66",
+    name = "Justfile",
+  },
   [".luaurc"] = {
     icon = "",
     color = "#00a2ff",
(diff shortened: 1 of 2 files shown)


lua/nvim-web-devicons/icons-default.lua
```diff
@@ -107,6 +107,12 @@ local icons_by_filename = {
     cterm_color = "28",
     name = "Gvimrc",
   },
+  [".justfile"] = {
+    icon = "",
```
alex-courtis, line 111: Is that a just specific icon?

If not can we please use the general `"config"` icon ``

chrisgrieser: I simply used the icon that `eza` is using for Justfiles. That airtight it should I change it?

alex-courtis: Is it [eza](https://github.com/eza-community/eza) you are referring you to?

If so, how do we see their icon (source)?

chrisgrieser: Yes. I think this is the location in their source code: https://github.com/eza-community/eza/blob/main/src/output/icons.rs#L281C4-L281C57

and in the terminal, shows like this: 
Pasted image 2024-06-07 at 02 01 04@2x

alex-courtis (approved):

chrisgrieser:
> Wow... that looks very similar. We should probably sync up at some point.

That would indeed be nice for consistency between nvim and the terminal. Maybe establish a common "standard" for matching filetypes with nerdfont icons? eza and devicons cannot be the only ones who use those icons for filetypes. 

Anyway, that's a different topic, thanks for the merge.

PR #480, opened 2024-07-13
fix: broken .prettierrc icon

xuanhung1509:

PR #481, opened 2024-07-13
fix: broken .prettierrc icon (light)

xuanhung1509:


alex-courtis:
#480 covers this change

PR #484, opened 2024-07-22
feat: add apl bqn

Tortus-exe:
This pull request adds icons for the APL and BQN filetypes. That's it. 
For regular users of the language, this will be a nice touch.


alex-courtis (approved):

PR #485, opened 2024-07-23
fix(gleam): move from icons_by_filename to icons_by_file_extension

Massolari:
According to the documentation
> There are two tables where icons can be added:
> 1. icons_by_filename
> 2. icons_by_file_extension
>
> Add the icon in table 1. if the icon is for a file that is always named that
> way, for example `.gitconfig`. Add to table 2. if the icon is for all files
> with an extension.

the Gleam icon must be placed under `icons_by_file_extension`

Disclaimer: it was not working for me and that's the reason

### DIFF
--- a/lua/nvim-web-devicons/icons-default.lua
+++ b/lua/nvim-web-devicons/icons-default.lua
@@ -101,12 +101,6 @@ local icons_by_filename = {
     cterm_color = "196",
     name = "GitModules",
   },
-  [".gleam"] = {
-    icon = "",
-    color = "#ffaff3",
-    cterm_color = "219",
-    name = "Gleam",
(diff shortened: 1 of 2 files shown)


alex-courtis (changes requested):
Let's leave it in `icons_by_filename` please; it may break some users...

Massolari:
I can do that.
But it doesn't make sense. In Gleam, there is no file called `.gleam`

alex-courtis:
> I can do that. But it doesn't make sense. In Gleam, there is no file called `.gleam`

Please accept my apologies; I misread as filetype...

alex-courtis (approved):

PR #495, opened 2024-09-12
feat: add .prettierrc.{cjs,js,mjs}

ttytm:
Includes missing entries for the prettier config. 

Stumbled over it while using astro which defaults to `.prettierrc.mjs`


hasecilu (approved):
LGTM

image

PR #496, opened 2024-09-15
feat: add variant="light|dark" option to override &background

williamhCode:
PR for the discussion here: https://github.com/nvim-tree/nvim-web-devicons/discussions/491#discussioncomment-10649821

I have a few questions about the code. Right now, `refresh_icons()` is called directly in the `nvim-web-devicons.lua` file. However, now that I'm adding an option which the `refresh_icons()` function depends on, should I move the call into the `setup()` function? Right now im just calling `refresh_icons()` again if the variant option is not "auto", which is a bit inefficient.

### DIFF
--- a/README.md
+++ b/README.md
@@ -77,7 +77,7 @@ require'nvim-web-devicons'.setup {
  -- because its name happened to match some extension (default to false)
  strict = true;
  -- set the light or dark variant manually, instead of relying on `background`
- -- (default to "auto")
+ -- (default to nil)
  variant = "light|dark";
  -- same as `override` but specifically for overrides by filename
  -- takes effect when `strict` is true
--- a/README.md
+++ b/README.md
@@ -36,7 +36,8 @@ Run `:NvimWebDeviconsHiTest` to see all icons and their highlighting.
 
 ### Variants
 
-Light or dark color variants of the icons depend on `&background`.
+Light or dark color variants of the icons depend on `&background`.  
+The variant can also be set manually in `setup` with the `variant` option.
 
 The variant is updated:
 - on `OptionSet` event for `background`, or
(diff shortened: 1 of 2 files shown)


alex-courtis:
> Right now im just calling `refresh_icons()` again if the variant option is not "auto", which is a bit inefficient.

That's unavoidable; we can't do anything until we know their choice.

Setup actually expects the icons to be populated i.e. refresh_icons has been called: https://github.com/williamhCode/nvim-web-devicons/blob/83238ce4809b4a34282ca8148e26e2d3fbf09ff7/lua/nvim-web-devicons.lua#L397

Suggestion: move the refresh_icons call at 595 to setup.

README.md
```diff
@@ -75,6 +76,9 @@ require'nvim-web-devicons'.setup {
  -- prevents cases when file doesn't have any extension but still gets some icon
  -- because its name happened to match some extension (default to false)
  strict = true;
+ -- set the light or dark variant manually, instead of relying on `background`
+ -- (default to "auto")
```
alex-courtis, line 80: `"auto"` is not really needed and will be a little confusing.

Let's remove it.

lua/nvim-web-devicons.lua
```diff
@@ -40,12 +40,19 @@ local global_opts = {
   strict = false,
   default = false,
   color_icons = true,
+  variant = "auto",
```
alex-courtis, line 43: Let's not use "auto" please, we don't really need it.

`variant` nil or any other value should be the else case in the if around line 49.

alex-courtis (changes requested):
Works beautifully, `:NvimWebDeviconsHiTest` shows correct colours and highlighting in all tests.

README.md
williamhCode: Changed to nil.

lua/nvim-web-devicons.lua
williamhCode: Changed to nil, fixed if else logic

williamhCode:
Additionally, i moved the refresh to right after setting variant global opt

alex-courtis (approved):
Simple and bulletproof, tested OK again.

williamhCode:
No problem!

PR #500, opened 2024-09-28
feat: add hyprpaper icons

pointless-existence:
Add's suport for hyprpaper (https://github.com/hyprwm/hyprpaper) config files, using the same config as the rest of the hypr suite.


hasecilu (approved):
LGTM
image

PR #501, opened 2024-09-28
feat: add ebuild file extension

hasecilu:
Gentoo Ebuild file

image
image


alex-courtis (approved):
Many thanks @hasecilu 

Interesting that the same colour works for light and dark.

PR #503, opened 2024-10-04
fix(#502): explicit case insensitive matching of filenames

### LINKED ISSUE #502: Override does not work
```
  return require("nvim-web-devicons").setup({override = {Jenkinsfile = {color = "#FF0000", cterm_color = "52", icon = "", name = "Jenkinsfile"}}}})
```

And here it is Jenkinsfile, for instance, showing the default icon:

image

alex-courtis:
fixes #502 

I'm always nervous about fixes like this... is this safe @hasecilu ?

### DIFF
--- a/lua/nvim-web-devicons/icons-default.lua
+++ b/lua/nvim-web-devicons/icons-default.lua
@@ -275,12 +275,6 @@ local icons_by_filename = {
     cterm_color = "28",
     name = "Vimrc",
   },
-  ["R"] = {
-    icon = "󰟔",
-    color = "#2266ba",
-    cterm_color = "25",
-    name = "R",
-  },
   ["avif"] = {
     icon = "",
     color = "#a074c4",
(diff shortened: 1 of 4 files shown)


hasecilu:
It seems it's working for the files you mentioned were not working.

image

lua/nvim-web-devicons/icons-default.lua
```diff
@@ -275,12 +275,6 @@ local icons_by_filename = {
     cterm_color = "28",
     name = "Vimrc",
   },
-  ["R"] = {
-    icon = "󰟔",
-    color = "#2266ba",
-    cterm_color = "25",
-    name = "R",
-  },
```
alex-courtis, line 283: This and `r` appear to be incorrectly placed extension.

alex-courtis:
Thanks mate.

PR #522, opened 2024-12-13
fix: update `vercel.json` glyph from Unicode to Nerd Fonts

hasecilu:
Nerd Font glyph is bigger.

2024-12-12-192551_hyprshot


alex-courtis (approved):
Nice find!

PR #525, opened 2024-12-14
fix: update Gemfile and valgrant files incorrect definitions

### LINKED ISSUE #510: Gemfile ruby icon is not used due to wrong definition
image
Gemfile ruby icon is not used due to wrong definition `"gemfile$"` here:
https://github.com/nvim-tree/nvim-web-devicons/blob/19d257cf889f79f4022163c3fbb5e08639077bd8/lua/nvim-web-devicons/icons-light.lua#L464C4-L464C14
```
  ["gemfile$"] = {
    icon = "",
    color = "#701516",
    cterm_color = "52",
    name = "Gemfile",
  },
```
However if i manually create new definition in my nvim config that uses `   ["Gemfile"] = {` instead of `["gemfile$"] = {` it works.

Santhoshmani1:
Fixes #510 

Updated the definitions of the `gemfile` and `vagrantfile` by removing the unused `$` sign. 

## Before fix
Screenshot 2024-12-14 185616

## After fix
Screenshot 2024-12-14 185356

### DIFF
--- a/lua/nvim-web-devicons/icons-default.lua
+++ b/lua/nvim-web-devicons/icons-default.lua
@@ -515,7 +515,7 @@ local icons_by_filename = {
     cterm_color = "160",
     name = "FreeCADConfig",
   },
-  ["gemfile$"] = {
+  ["Gemfile"] = {
     icon = "",
     color = "#701516",
     cterm_color = "52",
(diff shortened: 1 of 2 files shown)


hasecilu:
Are really the $ not needed?

Santhoshmani1:
They seems to be the issue in matching the files. I also tested by modifying these for my existing config under ~/local/share directory.

hasecilu (approved):
OK, I checked against `eza` and they don't have the "$", I don't know why it was added here.

PR #526, opened 2024-12-16
fix(#502): always use lowercase icons by filename

alex-courtis:
After looking at #525 it seems that #502 has regressed - lowercased filenames are not used.

Applied the lowercasing duing refresh.

Much nicer:


lua/nvim-web-devicons.lua
```diff
@@ -91,9 +116,9 @@ local filetypes = {
   ["checkhealth"] = "checkhealth",
   ["commit"] = "commit_editmsg",
   ["copying"] = "copying",
-  ["gemfile"] = "gemfile$",
+  ["gemfile"] = "gemfile",
   ["lesser"] = "copying.lesser",
-  ["vagrantfile"] = "vagrantfile$",
+  ["vagrantfile"] = "vagrantfile",
```
alex-courtis, line 121: The `$` had no purpose and would not match. There are no more.

hasecilu: I think that some programs use it for backup purposes, similar with `~` character

hasecilu (approved):
Looks good

lua/nvim-web-devicons.lua
alex-courtis: Oh I see. I thought it was a regex.

I'll back them out.

PR #530, opened 2024-12-23
feat: add icons for Assembly files (*.S and *.asm)

shizeeg:


hasecilu:
Hi, you need to move to the `icons_by_file_extension` table

shizeeg:
oh, my bad. (fixed)
the thing is `*.S` files are GNU and `*.asm` files are usually nasm so they are not the same (kinda).

I'm not sure myself which one to use so I put them both. Which one would it be, then 'ASM' or the 'chip' one?

hasecilu:
Not sure how much changes between different architectures, personally I immediately thought of micrcontrollers.

image




> P.S. what about `nvim-web-devicons.lua`? Your guidelines say I have to add new icons to

Not all types of files have a defined "filetype" on neovim, but yeah should be checked. In this case seems to be a `asm.vim` but not s.
I checked with `ls /usr/share/nvim/runtime/ftplugin/`

image

shizeeg:
both `code.S`  (GNU Assembler) and `code.asm` (nasm, fasm, etc.) are recognized  by NeoVim so I'm not sure what you meant. Since you mentioned 'ASM' icon is easier to read, I've replaced the 'chip' one with it.

hasecilu:
You need to change the "S" to lowercase. Also add `asm` to filetype, 

Regarding other filetypes I made a PR #531 to fix the other ones.

shizeeg:
isn't it supposed to be uppercase? Here for example https://github.com/torvalds/linux/blob/master/arch/x86/boot/copy.S

I thought it's a weird `gas` naming convention. Whatever I did as you said.

hasecilu:
Yes, it should be uppercase but it wasn't working that way, now it works.

hasecilu (approved):

PR #433, opened 2024-04-03
feat: add many icons

hasecilu:
Still WIP I'm looking at more icons, I want to hear some ideas.
~~Also I'm waiting for new Nerd Fonts release that includes some new icons.~~ 3.2.0 ver. released, I included the KDE, posmarketOS & Qt logos that are new.

https://github.com/ryanoasis/nerd-fonts/releases/tag/v3.2.0

---

For example in:
```lua
local filetypes = {
  ["avif"] = "avif",
  ["bzl"] = "bzl",
...
}
```
_Would be possible to create groups, for example create a single icon entry and use it various times for different file extensions to simplify the code?_ For example the open PR with music extensions, maybe a group of video, compressed files. In my case in this PR check the FreeCAD and KiCad extensions.

Something like:
```lua
local filetypes = {
...
  ["fcmacro"] = "freecad",
  ["fcmat"] = "freecad",
  ["fcparam"] = "freecad",
  ["fcstd"] = "freecad",
...
}
```

PR #531, opened 2024-12-24
fix: add missing filetypes

hasecilu:
Based on the script: https://github.com/nvim-tree/nvim-web-devicons/pull/433#issuecomment-2044188895


gegoune:
Nice! Might be worth adding to required checks on CI.

scripts/filetypes.sh
```diff
@@ -0,0 +1,25 @@
...
+	fi
+
+	if $in_section && [[ $line =~ \[\"(.+)\"\] ]]; then
+		key="${BASH_REMATCH[1]}"
+		# Search for the key in the main file
+		line=$(grep -F "\"$key\"" "lua/nvim-web-devicons.lua")
+		if [ -n "$line" ]; then
+			continue
+		else
+			[ -f "/usr/share/nvim/runtime/syntax/$key.vim" ] &&
```
hasecilu, line 21: This is valid on Arch Linux but is it the same path on Ubuntu (CI)?

CONTRIBUTING.md
```diff
@@ -4,22 +4,24 @@ Thank you for your contribution!
 
 ## Order
 
-Please ensure `icons_by_filename`, `icons_by_file_extension` and `filetypes` are ordered alphabetically, to prevent merge conflicts.
+Please ensure `icons_by_filename`, `icons_by_file_extension`, `icons_by_operating_system`, `icons_by_desktop_environment`,
```
hasecilu, line 7: Line breaks at 120 char but could be 100 or 80

alex-courtis:
> Nice! Might be worth adding to required checks on CI.

Yes please!

CONTRIBUTING.md
```diff
@@ -46,15 +51,20 @@ make style-fix
...
+
+1. `icons_by_filename`
+2. `icons_by_file_extension`
+3. `icons_by_operating_system`
+4. `icons_by_desktop_environment`
+5. `icons_by_window_manager`
+
+Add the icon to table **1.** if the icon is for a file that is always named that way, for example `.gitconfig`.
+Add the icon to table **2.** if the icon is for all files with an extension, for example `vim`.
+Add the icon to table **3.**, **4.**, **5.** if the icon is from an OS, DE or WM.
```
alex-courtis, line 64: Thank you!

scripts/filetypes.sh
```diff
@@ -0,0 +1,25 @@
+#!/usr/bin/env bash
```
alex-courtis, line 1: Shellcheck OK
Looks like we do need bash (not /bin/sh) for the pattern matching.

CONTRIBUTING.md
alex-courtis: I'm easy either way, although links often end up >80

alex-courtis (changes requested):
Fantastic! Working as advertised.

Please:
- [ ] delete `scripts/filetype-generator.sh`
- [ ] add to CI

alex-courtis:
Dumb question 1:

Could we generate these mappings at runtime, rather than compile time? Users may have installed other filetypes.

The mapping would still need to exist, for unconventional mappings like `["commit"] = "commit_editmsg",`

It looks like we could use [vim.fn.getcompletion("", "filetype", false)](https://neovim.io/doc/user/builtin.html#getcompletion()) at startup.

alex-courtis:
Dumb question 2:

edit: never mind, I didn't read the doc

hasecilu:
> Could we generate these mappings at runtime, rather than compile time? Users may have installed other filetypes.

Or let users add new filetypes on the config file so if they have more filetypes they know they have to also include it.

alex-courtis:
> Or let users add new filetypes on the config file so if they have more filetypes they know they have to also include it.

Yes, that is necessary and we have many cases.

I'm happy whichever way you go.

alex-courtis:
Looking good, I created a branch to test: #534
It removed `asm` to test.

The script doesn't fail when a file type is missing:

```
: ; make filetypes
./scripts/filetypes.sh
Please add "asm" to filetypes Lua table.
: ; echo $?
0
: ; scripts/filetypes.sh
Please add "asm" to filetypes Lua table.
: ; echo $?
0
: ;

It's going to need to return non-zero (1 will do) when a type is missing, so that CI can fail.

alex-courtis:
It looks like `$VIMRUNTIME` is not set in CI; we're going to need to set it as per https://github.com/nvim-tree/nvim-tree.lua/blob/5ad87620ec9d1190d15c88171a3f0122bc16b0fe/.github/workflows/ci.yml#L91

I've instrumented the script to show what's present: https://github.com/nvim-tree/nvim-web-devicons/actions/runs/12539821919/job/34966371519?pr=535

It looks like we can use `/home/runner/nvim-stable/share/nvim/runtime`

alex-courtis:
Beautiful!

https://github.com/nvim-tree/nvim-web-devicons/actions/runs/12590541237/job/35092246843?pr=539

alex-courtis (approved):

alex-courtis:
Added filetypes to required CI runs

hasecilu:
Nice! Glad it worked

PR #533, opened 2024-12-29
Add missing dockerfile and dockercompose icons

prashantdivate:
After opening Nerdtree with latest nvim version I was not seeing dockerfile and dockercompose icons maybe older icons were incompatible with latest version of nerdfonts

### DIFF
--- a/lua/nvim-web-devicons/icons-default.lua
+++ b/lua/nvim-web-devicons/icons-default.lua
@@ -462,14 +462,14 @@ local icons_by_filename = {
     name = "Dockerfile",
   },
   ["docker-compose.yml"] = {
-    icon = "󰡨",
-    color = "#458ee6",
+    icon = "",
+    color = "#0db7ed",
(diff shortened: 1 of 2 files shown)


prashantdivate:
FYI,
image

hasecilu:
That's odd, icons show in my end, are you sure your font is OK? Let's wait is someone else have also same problem.

gegoune:
> That's odd, icons show in my end, are you sure your font is OK? Let's wait if someone else have same problem.

Same here, works just fine.

alex-courtis:
Could it be your nerd font version?

3.3.0 is the latest.

prashantdivate:
I have fonts installed and running `fc-cache -fv` as well shown below
image

Am I missing something then ?

hasecilu:
Try `fc-list :charset=f0868` to check which font provides the docker glyph. Also check the font name is properly set on your terminal config.

PR #543, opened 2025-01-06
chore: sort icons alphabetically

gegoune:
Closes #319

TODO:
- [x] discuss formatting
- [x] add CI validation
- [x] adapt `generate_colors.lua`
- [x] refactor `generate_colors.lua`

What's the consensus for changing format to single line tables like that? This makes sorting alphabetically much easier. We would probably have to find a way to enforce that formatting on CI.

PR #550, opened 2025-01-25
refactor: split tables into individual files

gegoune:


scripts/generate_colors.lua
```diff
@@ -104,23 +115,21 @@ end
...
-io.write "Generating colors for light background..."
-
-iterate_colors(generate_for_light_bg)
-
-vim.cmd "silent! wall!"
-io.write " OK\n"
+for _, file in ipairs(_G.ICON_FILES) do
+  vim.cmd("noswapfile drop lua/nvim-web-devicons/light/" .. file)
+  io.write("Generating colors for light background: " .. file .. "...")
+  iterate_colors(generate_for_light_bg)
```
alex-courtis, line 132:
```suggestion
  iterate_colors(generate_for_light_bg)
    vim.cmd(
    string.format(
      "1s/.*/&      -- this file is generated from lua\\/nvim-web-devicons\\/default\\/%s, please do not edit",
      file
    )
  )
```

alex-courtis (changes requested):
Works beautifully, removes more code.
Please:

- [ ] Update CONTRIBUTING.md
- [x] change the operation name from colors to generate

Makefile
alex-courtis, line 1: More phony targets now:
```
.PHONY: all clean colors colors-check vim-colortemplate mini-align style-check style-fix lint filetypes
```

alex-courtis:
Whoops, I see you're still changing. I'll wait until you're done :)

scripts/generate_colors.lua
gegoune: Neat. Will add now. Do we also want to let GH know about it?
https://docs.github.com/en/repositories/working-with-files/managing-files/customizing-how-changed-files-appear-on-github

Makefile
gegoune: Don't think we want to have `vim-colortemplate` and `mini-align` as phony, they create files and with them declared as phony both plugins get re-dowloaded on each target run.

alex-courtis: Whoops... forgot about directories...

scripts/generate_colors.lua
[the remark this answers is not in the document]
gegoune: Would just `generate.lua` work for you?

gegoune:
Sorry for pushing so vigorously to ready for review PR. To stop myself from squeezing even more into this poor PR I would like to ask yo to review once again.

On top of what you already commented on:
- do we want to break off `filetypes` table to separate file? My reasoning was that it is one of those 'PR entry points' for new icons, which contributors could benefit from having easier access to.
- types: I only have very superficial knowledge on the matter

Will update contributing document next.

alex-courtis:
> * do we want to break off `filetypes` table to separate file? My reasoning was that it is one of those 'PR entry points' for new icons, which contributors could benefit from having easier access to.

Absolutely! It's a great source of confusion, and data doesn't belong next to code.

alex-courtis:
> * types: I only have very superficial knowledge on the matter

Looks good to me; luals will complain if they're not right.

There are two warnings right now for deprecated... I'll deal with them.

CONTRIBUTING.md
```diff
@@ -24,8 +19,43 @@ or via your OS package manager e.g. *Arch Linux*:
 pacman -S stylua luacheck
 ```
 
+## Adding icons
```
gegoune, line 22: Note: I didn't just move the block here, so please review as new.

.gitattributes
```diff
@@ -0,0 +1,2 @@
+lua/nvim-web-devicons/default/* linguist-generated=true
```
gegoune, line 1: This shouldn't be here. 
```suggestion
```

.gitattributes
```diff
@@ -0,0 +1 @@
+lua/nvim-web-devicons/light/* linguist-generated=true
```
hasecilu, line 1: So, this helps to simplify diffs on PR?

CONTRIBUTING.md
hasecilu: LGTM

.gitattributes
gegoune: Correct. By default those will be hidden with message that generated files are hidden by default.

hasecilu: Cool, new trick learned.

hasecilu (approved):
Working flawlessly!

scripts/generate_colors.lua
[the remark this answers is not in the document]
alex-courtis: Yes please.

lua/nvim-web-devicons.lua
```diff
@@ -1,9 +1,20 @@
 local M = {}
 
+---@alias iconName string Name of the icon
+
+---@class Icon
```
alex-courtis, line 5: This works. It will warn when you do:

```lua
---@param icon_data Icon
local function get_highlight_foreground(icon_data)
```

alex-courtis:
High level testing:

- [x] filetypes happy path - exit 0
- [x] filetypes missing - message, exit 2
- [x] filetypes bad - exit 2
- [x] filetypes bad and missing - message, exit 2
- [x] generate happy path - exit 0
- [x] generate with bad default file - works but exit 2
- [x] generate with bad light file - no errors and exit 0
- [x] generate with missing light file - no errors and exit 0
- [x] generate with no light files - fails with exit 2, but message is meaningful
- [x] align happy path - works
- [x] align bad default - works, exit 2
- [x] runs as normal

Makefile
alex-courtis, line 1: These are all present in CI and work.

CI could be consolidated at some point but not today.

gegoune:
We can harden it. Which cases where scripts worked for invalid inputs are worth fixing in your opinion?

CONTRIBUTING.md
alex-courtis: Looks good, reverted contributing changes on #551

alex-courtis (approved):
This is a huge win and will save a lot of review work.

alex-courtis:
> We can harden it. Which cases where scripts worked for invalid inputs are worth fixing in your opinion?

I reckon it's fine - invalid inputs are obvious and will fail CI.

> > generate with no light files - fails with exit 2, but message is meaningful
> 
> Seen that. Adding force to rm fixes the issue, will push now.

Nice.

> 
> > generate with bad default file - works but exit 2
> 
> This is probably most worrisome. I would like to enforce single line icon tables somehow, ideally with external formatter rather than coming up with potentially fragile solution ourselves. Any ideas? You did mention emmylua in [#543 (comment)](https://github.com/nvim-tree/nvim-web-devicons/pull/543#issuecomment-2586011922) but I am not sure if `align_continuous_similar_call_args` what we are after, but didn't check beyond having a quick glance.

Honestly, garbage in, garbage out.

I don't think it's worth it to add/move to emmylua. It's not going to handle cases like this either - it will just format them the same as mini does.
mini is doing `align_continuous_similar_call_args` so nothing to do here.

gegoune:
Thanks for reviewing, both!

PR #559, opened 2025-01-27
fix: add missing filetype icon mapping for Norg

ALVAROPING1:


gegoune (approved):

hasecilu:
Do you have a Syntax Vim file for Norg on your system?

gegoune:
There wasn't one on nightly, but there is mapping for it in `filetype.lua`.

hasecilu:
Ok, so potentially we should also look into this file for other missing extensions?

gegoune:
Indeed, as I mentioned in #560.

hasecilu (approved):

alex-courtis (approved):

Issue #540, opened 2025-01-06
FR: user should be able to specify colours for both light and dark backgrounds


gegoune:
It is currently impossible(?) to set separate colours for light and dark background, in the same way vein built in icons change their colours based on background.

PR #562, opened 2025-02-27
feat: let users override icons

hasecilu:
Previously if you wanted to change one icon across several
extensions/filenames you needed to manually override each
instance, now you can target specific icons to simplify
config file.

Basically I'm implementing this feature to match icons to
eza and lazygit, and simplify setup to avoid big config file.

### config file used

| Before | After |
|--------|--------|
| image | image | 

### config file used

```lua
return {
  "nvim-tree/nvim-web-devicons",
  event = "VimEnter",
  opts = {
    override_by_icon = {
      [""] = { icon = "" },
      [""] = { icon = "" },
    },
  },
}
```

Probably it's also a good idea to:
- [ ] override color

### DIFF
--- a/lua/nvim-web-devicons.lua
+++ b/lua/nvim-web-devicons.lua
@@ -239,6 +239,7 @@ function M.setup(opts)
   local user_operating_system_icons = user_icons.override_by_operating_system
   local user_desktop_environment_icons = user_icons.override_by_desktop_environment
   local user_window_manager_icons = user_icons.override_by_window_manager
+  local user_icon_overrides = user_icons.override_by_icon
 
   -- filename matches are case insensitive
   lowercase_keys(icons_by_filename)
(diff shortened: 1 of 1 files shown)


alex-courtis:
That's rather nice... we should have done this years ago.

My initial thought was dark/light variants, but that's a bigger task #540

lua/nvim-web-devicons.lua
```diff
@@ -356,6 +359,11 @@ local function get_icon_data(name, ext, opts)
     icon_data = icons[name] or get_icon_by_extension(name, ext, opts) or (has_default and default_icon)
   end
 
+  -- Check for icon override
+  if icon_data and global_opts.override[icon_data.icon] then
+    icon_data = vim.tbl_extend("force", icon_data, global_opts.override[icon_data.icon])
+  end
```
alex-courtis, line 365: A bit slow but the safety is worth it.

hasecilu: I tried also changing color but didn't work, so probably will be better to first implement better color overrides and thee try again.

alex-courtis:
Is this good to go @hasecilu ? Can we merge?

alex-courtis:
Closing due to inactivity.

It would be great if we could continue work on this one.

PR #563, opened 2025-03-02
feat: add bun.lock icon

artrixdotdev:
Bun has recently added a text lock file called `bun.lock`, and its now the default for new bun projects. Probably a good idea to add an icon for it too.


alex-courtis (approved):

Please run `make` when raising a PR as per [Contributing](https://github.com/nvim-tree/nvim-web-devicons/blob/master/CONTRIBUTING.md) to format and generate light variants.

In the interest of speed I've done this and pushed.

Issue #572, opened 2025-04-06
Bug: user icons set from setup are lost after refresh


dsaenztagarro:
## First some words...

Hi, first of all, I appreciate a lot this open source plugin, and really, I'm super happy to have a chance to bring some gratitude for all the work from the NeoVim community. That said! Here is the issue...

## Problem
I've realized that this plugin is not applying the icons set by the user by calling to `setup` function (e.g. icons set from `overrides_by_extension`.

I've done some investigation, and I have found that in https://github.com/nvim-tree/nvim-web-devicons/pull/270 pull request, we introduced this auto command, in order to change icons depending on the background:

https://github.com/nvim-tree/nvim-web-devicons/blob/4c3a5848ee0b09ecdea73adcd2a689190aeb728c/lua/nvim-web-devicons.lua#L453-L456

From the issue description [here](https://github.com/nvim-tree/nvim-web-devicons/issues/226), I read:

> I have no config for nvim-web-devicons.

What it makes me think that the functionality was added without verifying that user overrides were kept.

It happens that NeoVim sets the background on startup automatically, as you can see in `:help background`

![Image](https://github.com/user-attachments/assets/8bc70409-a5e0-462b-8b26-5f2a39b14f45)

So just after all plugins are loaded with Lazy, the auto command is executed by NeoVim, loosing all user overrides.

**As result, user icon overrides are not applied ever.**

## Requirement
Apply changes needed so user's icons overrides are kept on background changes (i.e. also applied just after opening NeoVim).

P.D: Hope all this makes sense to you 🙏

alex-courtis:
Reopened following #575

dsaenztagarro:
@alex-courtis just in case, adding a comment here, ready for you to review! Thanks in advance! 😅

PR #573, opened 2025-04-06
fix: ensure icon overrides apply after refresh

dsaenztagarro:
Resolves https://github.com/nvim-tree/nvim-web-devicons/issues/572

### What I did

Git diff maybe result a little bit confusing, but I only extracted `apply_user_icons()` from `setup()`, so it can be called also later during the `refresh` callback.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -21,7 +21,7 @@ repos:
       - id: colors
         name: colors
         description: Ensures Light Color Scheme version has been generated.
-        entry: make colors
+        entry: make colors-check
         language: system
         require_serial: true
         pass_filenames: false
--- a/lua/nvim-web-devicons/filetypes.lua
+++ b/lua/nvim-web-devicons/filetypes.lua
@@ -132,6 +132,7 @@ return {
   ["odin"] = "odin",
   ["openscad"] = "scad",
   ["opus"] = "opus",
+  ["org"] = "org",
   ["otf"] = "otf",
   ["pck"] = "pck",
   ["pdf"] = "pdf",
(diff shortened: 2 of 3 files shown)


.pre-commit-config.yaml
```diff
@@ -21,7 +21,7 @@ repos:
       - id: colors
         name: colors
         description: Ensures Light Color Scheme version has been generated.
-        entry: make colors
+        entry: make colors-check
```
alex-courtis, line 24: Thank you

alex-courtis (approved):
Unwound the diff, changes are simply:
* `apply_user_icons` just moves the code out of setup unchanged
* `set_icon` changes param to `user_icons_opts` to disambiguate

I wasn't able to replicate the issue, however changes are applied as expected on manual refresh and autocmd.

lua/nvim-web-devicons/filetypes.lua
```diff
@@ -132,6 +132,7 @@ return {
   ["odin"] = "odin",
   ["openscad"] = "scad",
   ["opus"] = "opus",
+  ["org"] = "org",
```
alex-courtis, line 135: Unrelated nvim upstream addition. Not worth its own PR.

dsaenztagarro:
@alex-courtis I've noticed you have reverted the change after being merged.. I feel confused.. can you please explain to me the reason? 

The auto command that refresh the icons doesn't take care of keeping the icons overridden from the plugin setup configuration, and without that change, I won't be able to see the icons that I was expecting :(

I would appreciate a lot if you find the time for a few words, so I can see if I can bring to you the explanations or examples needed to keep up with the change.

Thanks in advance

dsaenztagarro:
@alex-courtis I see! I totally understand the reasons and I completely agree with your criteria. Sure, I will review with care on the following days.

PR #600, opened 2025-08-22
feat: add frag, geom, glsl, vert icons

Natzgun:
Icons for OpenGL pipeline


alex-courtis (changes requested):
Could we please use the `nf-dev-opengl` icon `` ? https://www.nerdfonts.com/cheat-sheet

It has an official icon and colour: https://opengl.org/

alex-courtis (approved):

PR #621, opened 2026-02-21
feat: add dune icon

ashton:
Adding icon for `dune` files. [Dune](https://dune.build/) is a build tool for Ocaml projects.

Since there is no icon for dune in nerd fonts, I used `nf-dev-akka` which is the Scala Akka icon, but it seems a lot like a dune to me 😂  not sure if this is a problem.

Here's a SC on how it looked like in a OCaml project.

dune-webdev-icon


alex-courtis (changes requested):

The icon does indeed look like a dune, however it is for a different framework that we may one day add.

Proposal:
Use the ocaml icon, with a different colour. We have followed that pattern for python and other frameworks/languages.

Colour is up to you :)

alex-courtis (approved):
Your additions are most gratefully appreciated.

PR #625, opened 2026-04-01
feat: add .clangd icon

antinomie8:
Strangely enough, there is both `.clang-tidy` and `.clang-format` icon but no `.clangd` one, so I added it (and added two unrelated commits to fix a typo and remove useless newlines).


alex-courtis:
> In case you want to investigate what causes the issues in `generate.lua` when using 0.12, here are the errors I got:

I'm not able to replicate this one.

Which package of neovim are you using? I see it's in `/usr/local/share`

I've been testing successfully by installing the stable tag from source and setting $VIMRUNTIME `/usr/local/share/nvim/runtime`

PR #629, opened 2026-04-07
fix: update icons for yaml/yml extensions

fjfnaranjo:


alex-courtis (changes requested):

Please:

- [x] Update the light icon colours etc as per [CONTRIBUTING.md](https://github.com/nvim-tree/nvim-web-devicons/blob/master/CONTRIBUTING.md#building)

Should we consider changing the colour to `#cc1018` as per https://yaml.org/assets/logo.png

alex-courtis:
> But I reviewed it carefully and I went with cterm Red3, instead of Red1. Is less saturated and should render ok.

Looks good, blends in with others and won't stand out too much.

alex-courtis (changes requested):
We have a [build failure](https://github.com/nvim-tree/nvim-web-devicons/actions/runs/24130601022/job/70518797198?pr=629) - the cterm icons are being overwritten.

We need to go the other way - set the `color` and `cterm_color` will be updated in all places when you build.

I've updated the instructions for more clarity: https://github.com/nvim-tree/nvim-web-devicons?tab=contributing-ov-file#adding-or-updating-icons

fjfnaranjo:
Oh. I see. Thanks!

I guessed the "light" icons was a file intended to supply an alternative design/color for non-dark backgrounds. Looks like its completely automated and uses the values from the other file (I didn't see the comment in the top of the file).

Also, I wasn't running the make targets, I was editing directly. I guess that is the main issue for my PRs failing in the CI. The "building" is like a "pre-push" requirement, as if it was a regular program.

What I was doing is using my local Neovim install to verify it worked correctly.

I will prepare a container to build it and send a new commit when it does without failures.

fjfnaranjo:
I did the `make` this time, having modified only the "default" file.

I'm leaving here a Debian 13 based container definition to run `make`.

https://github.com/nvim-tree/nvim-web-devicons/pull/632

alex-courtis (approved):

PR #641, opened 2026-05-18
fix: use `vim.nonnil` if available

MariaSolOs:
https://github.com/neovim/neovim/pull/39495 deprecated `vim.F.if_nil`, so let's use the new API when available to avoid the deprecation warnings.


alex-courtis (approved):

PR #648, opened 2026-07-01
feat: add asmdef icon

cpea2506:
Add `asmdef` icon which is used by Unity Engine for Assembly Definition.

## Unity

image

## Neovim

image


gegoune:
Thanks!

PR #649, opened 2026-07-08
feat: Add icons by pattern and better env support

rokbot:
- This allows to use icons by lua patterns
- Improve icons for env specific files .env.example, .env.local etc.

Useful links:
 https://www.lua.org/pil/20.2.html
 https://www.lua.org/manual/5.1/manual.html#5.4.1
 https://gitspartv.github.io/lua-patterns/

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -70,13 +70,15 @@ Icons must be added to `/default/` files only. The `/light/` files are automatic
 
 1. `icons_by_filename.lua`
 2. `icons_by_file_extension.lua`
-3. `icons_by_operating_system.lua`
-4. `icons_by_desktop_environment.lua`
-5. `icons_by_window_manager.lua`
+3. `icons_by_pattern.lua`
+4. `icons_by_operating_system.lua`
+5. `icons_by_desktop_environment.lua`
+6. `icons_by_window_manager.lua`
 
 Add the icon to table in file **1.** if the icon is for a file that is always named that way, for example `.gitconfig`.
 Add the icon to table in file **2.** if the icon is for all files with an extension, for example `vim`.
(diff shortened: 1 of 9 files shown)


alex-courtis (changes requested):
# Questions

1. Why do we need to sort the user pattern icons into `user_patterns_sorted`? What is the significance of the sort?
2. Should this work with file names / extensions? I could only make it work with file types.

# Feedback

Thank you, this is a great addition to nvim-web-devicons.

I do have some concerns:

- Performance: pattern matching is expensive and web-devicons is used in a great many places. It must be as fast as possible.
  - Ameliorate: only use patterns for user overrides. The user can choose to incur the cost.
- Completeness: it seems we can only override filetypes, not names or extensions. Users _will_ want this. My first test was with a file name, however that did not work.
  - Proposal: remove `override_by_pattern` and allow users to specify patterns for all `override_by_`

Please reply with your thoughts and we can decide on the final direction.

# Work Needed, As Submitted

Please:

- [ ] Change from lua patterns to vim regex patterns to be processed by `vim.fn.match`. nvim-tree uses vim regex, as do most other plugins, and we should provide a consistent user experience.
- [ ] Explicitly specify which regexes are used, in the readme
- [ ] Resolve the style issue https://github.com/nvim-tree/nvim-web-devicons/actions/runs/29295576657/job/86978468164?pr=649

# Manual Testing

## Override Working

```lua
require'nvim-web-devicons'.setup {
  override_by_pattern = {
    ["^kconf.*$"] = {
      icon = "K",
      color = "#FF0000",
      name = "kconfsomething"
    }
  }
}
```
Shows icon and:
```
Overrides
---------
K  ^kconf.*$  DevIconKsomething  guifg=#ff0000
```

## Sample Working

`.env.foo` shows icon as expected

PR #651, opened 2026-07-13
docs: remove non-vim.pack installation methods

gegoune:
@alex-courtis Should we also remove luarocks badge?


alex-courtis:
> @alex-courtis Should we also remove luarocks badge?

Yes please... luarocks never really worked out. That release action has never even run.

We should to the same for the tree...

alex-courtis (approved):

PR #654, opened 2026-08-26
feat: change clang icons to llvm dragon logo

Pistonight:
Change clang config icons from the generic config icon to llvm logo, i picked the color from the colorized icon on llvm.org that has a good contrast (the wings are a bit too dark)


alex-courtis (approved):
Many thanks for your contribution, colours look good.

PR #655, opened 2026-09-16
feat: add wgsl icon

FineFindus:
Ref: https://www.w3.org/TR/WGSL/


alex-courtis (changes requested):
Thanks for the new icon.

It looks like there's actually an official icon at https://www.nerdfonts.com/cheat-sheet

Search for webgpu

FineFindus:
Oh, I think that's the only term that I didn't actually search :sweat_smile: 

Also updated the color to use a color from the WebGPU logo.

alex-courtis (approved):
Very nice, thank you.

PR #656, opened 2026-09-19
feat: update Elixir icon colors to the official brand palette

georgeguimaraes:
Elixir now publishes its brand palette on the trademarks page: https://elixir-lang.org/trademarks/. This updates the six Elixir entries (ex, exs, eex, heex, leex, mix.lock) from `#A074C4`, which came from the seti palette rather than the project, to `#B88CD9` (purple-40 from the logo palette, picked for contrast on dark backgrounds).

Light variants and cterm colors regenerated with `make`, `make colors-check` passes.


alex-courtis (approved):
Many thanks for the diligent contribution.

